PR #4, opened 2015-11-12
Separate "Save" & "Publish"

valendesigns:
Only show a “Save Draft” button if the user does not have the capability to `customize_publish`.


valendesigns:
@westonruter Do you have time for a code review?

westonruter:
This is in my todo list.

php/class-customize-snapshot-manager.php
```diff
@@ -131,6 +119,25 @@ public function __construct( Plugin $plugin ) {
 	}
 
 	/**
+	 * Set the Customizer return URL.
+	 */
+	public function set_return_url() {
+		global $wp_version;
+		if (
+			version_compare( $wp_version, '4.4-beta', '>=' )
```
westonruter, line 127: Good call :+1:

php/class-customize-snapshot-manager.php
```diff
@@ -103,23 +103,11 @@ public function __construct( Plugin $plugin ) {
...
-				$that->customize_manager->set_return_url( $return_url );
-			} );
-		}
-
+		add_action( 'customize_controls_init', array( $this, 'set_return_url' ) );
 		add_action( 'init', array( $this, 'maybe_force_redirect' ), 0 );
 		add_action( 'init', array( $this, 'create_post_type' ), 0 );
 		add_action( 'customize_controls_enqueue_scripts', array( $this, 'enqueue_scripts' ) );
-		add_action( 'wp_ajax_customize_save', array( $this, 'set_snapshot_uuid' ) );
+		add_action( 'wp_ajax_customize_save', array( $this, 'set_snapshot_uuid' ), 0 );
```
westonruter, line 110: This _might_ be better at `customize_save` priority 0 instead. I used `customize_save` action for Customize Setting Validation yesterday: https://github.com/xwp/wp-customize-setting-validation/blob/master/php/class-plugin.php#L40

In practice, the result would be identical right? But maybe `customize_save` would be less hack since it is an actual action that is mean to be hooked into?

php/class-plugin.php
```diff
@@ -37,6 +37,7 @@ public function __construct() {
 	function init() {
 		add_action( 'wp_default_scripts', array( $this, 'register_scripts' ), 11 );
 		add_action( 'wp_default_styles', array( $this, 'register_styles' ), 11 );
+		add_action( 'admin_init', array( $this, 'add_caps' ) );
```
westonruter: I _think_ that instead of using `admin_init` to add this role to the persistent role object that it could be better to just filter `user_has_cap` dynamically, for instance:

``` php
<?php
/* ... */

    add_filter( 'user_has_cap', array( $this, 'filter_user_has_cap' ), 9, 4 );
}

/* ... */

/**
 * Add the customize_publish capability to users who can edit_theme_options by default.
 *
 * @param array   $allcaps An array of all the user's capabilities.
 * @param array   $caps    Actual capabilities for meta capability.
 * @param array   $args    Optional parameters passed to has_cap(), typically object ID.
 * @param WP_User $user    The user object.
 */
function filter_user_has_cap( $allcaps, $caps, $args, $user ) {
    unset( $caps, $args, $user );
    if ( ! empty( $allcaps['edit_theme_options'] ) ) {
        $allcaps['customize_publish'] = true;
    }
    return $allcaps;
}
```

What do you think? This would mean the DB wouldn't have to be manipulated, and I recall from Nacin's talk a couple  years ago that filters are the best bet for adding caps to a role:

http://wordpress.tv/2013/08/10/andrew-nacin-current-user-can-watch-this-talk/
http://mannieschumpert.com/blog/wordpress-capabilities-magic-with-map_meta_cap/
https://gist.github.com/mannieschumpert/8886289

php/class-customize-snapshot-manager.php
valendesigns, line 110: Yes that could work, though `set_snapshot_uuid` would need to be modified to filter `customize_save_response` instead of retuning `wp_send_json_error`.

php/class-plugin.php
valendesigns: This could be simplified to

``` PHP
/**
 * Add the customize_publish capability to users who can edit_theme_options by default.
 *
 * @param array $allcaps An array of all the user's capabilities.
 */
function filter_user_has_cap( $allcaps ) {
    if ( ! empty( $allcaps['edit_theme_options'] ) ) {
        $allcaps['customize_publish'] = true;
    }
    return $allcaps;
}
```

PR #8, opened 2015-12-17
Add support for post widgets & widget subareas

valendesigns:


php/class-customize-snapshot.php
```diff
@@ -296,9 +296,45 @@ public function values() {
...
 				return $setting['dirty'] === $dirty;
 			} );
+
+			$customize_widgets = $this->snapshot_manager->customize_manager->widgets;
+			$dirty_widgets = array();
+
+			// Find widgets in the dirty sidebars.
+			foreach ( $dirty_values as $setting_id => $setting ) {
+				if ( -1 !== strpos( $setting_id, 'sidebars_widgets' ) ) {
+					$dirty_widgets += (array) $setting['value'];
```
westonruter: Using `+` for array concatenation yields unexpected results. I think you actually mean `array_merge()` here? See for example: https://3v4l.org/50WYF

westonruter: Maybe:

``` php
$dirty_widgets = array_unique( array_merge( $dirty_widgets, (array) $setting['value'] ) );
```

valendesigns: array_merge would work correctly

westonruter:
Closing this in favor of #9.

PR #11, opened 2016-03-18
v0.3.1

valendesigns:


westonruter:
Wow. @coveralls posts comments now?

valendesigns:
Oh that's really cool! I guess upgrading to 2.0 was worth it.

westonruter:
@valendesigns good on ya!

valendesigns:
@westonruter You want to review and merge this PR? I'll deploy to .org afterwards.

php/class-customize-snapshot-manager.php
```diff
@@ -90,14 +92,16 @@ public function __construct( Plugin $plugin ) {
 			$this->capture_unsanitized_snapshot_post_data();
 		}
 
-		$uuid = isset( $_REQUEST['customize_snapshot_uuid'] ) ? sanitize_key( wp_unslash( $_REQUEST['customize_snapshot_uuid'] ) ) : null; // WPCS: input var ok.
-		$scope = isset( $_REQUEST['scope'] ) ? sanitize_key( wp_unslash( $_REQUEST['scope'] ) ) : 'dirty';  // WPCS: input var ok.
+		$uuid = isset( $_REQUEST['customize_snapshot_uuid'] ) ? sanitize_text_field( sanitize_key( wp_unslash( $_REQUEST['customize_snapshot_uuid'] ) ) ) : null; // WPCS: input var ok.
+		$scope = isset( $_REQUEST['scope'] ) ? sanitize_text_field( sanitize_key( wp_unslash( $_REQUEST['scope'] ) ) ) : 'dirty';  // WPCS: input var ok.
```
westonruter, line 96: `sanitize_text_field()` is overkill here, since `sanitize_key()` is even more restrictive. But it doesn't hurt.

php/class-customize-snapshot.php
```diff
@@ -228,8 +228,10 @@ public function post( $refresh = false ) {
 			'posts_per_page' => 1,
 			'post_type' => Customize_Snapshot_Manager::POST_TYPE,
 			'post_status' => array( 'draft', 'publish' ),
+			'no_found_rows' => true,
+			'ignore_sticky_posts' => true,
```
westonruter, line 232: Could also add:

```
'update_post_meta_cache' => false,
'update_post_term_cache' => false,
```

Because neither terms nor postmeta will be used, right?

valendesigns, line 232: That is correct. However we could just do `'cache_results' => false`.

westonruter:
Looks good. Left a couple comments. Adding `sanitize_text_field()` in addition to `sanitize_key()` is overkill since the latter is more strict than the former. But it doesn't hurt.

Feel free to merge when you're ready.

php/class-customize-snapshot-manager.php
valendesigns, line 96: This is coming from VIP:

> use sanitize_text_field along with sanitize_key as sanitize_key alone is not meant for sanitizing user input

westonruter, line 96: OK :)

PR #23, opened 2016-04-27
Use wp_slash() instead of add_magic_quotes() since string val

westonruter:
`add_magic_quotes()` expects an array, and so it was doing nothing. The result was corruption of the JSON.


valendesigns:
@westonruter Should we release this as 0.3.2 or hold out for additional changes next week?

westonruter:
@valendesigns yes, we can release 0.3.2 with this and the other changes in `master` not yet released. I noticed no release tags exist for this plugin yet.

Issue #25, opened 2016-04-28
Store snapshot data in post_content instead of post_content_filtered


westonruter:
Follow pattern of Customize Widgets Plus in how data is stored for `widget_instance` posts.

valendesigns:
@westonruter I think this should be added to 0.4.0 so it's taken care of sooner rather than later since it has breaking changes.

Issue #12, opened 2016-04-20
Add UI for listing and accessing current snapshots


westonruter:
The custom post type for snapshots should be exposed in the admin so that the existing snapshots can be listed, with links to view them on the frontend or open them in the Customizer. Viewing the edit post screen for a snapshot could show a JSON dump of the settings in the snapshot.

valendesigns:
This still requires a link to view the Snapshot on the front-end before it can be closed.

westonruter:
Link for viewing snapshot on frontend: #49

PR #35, opened 2016-06-07
Release 0.4.0

valendesigns:
#### Pre Merge
- [x] Store snapshot data in post_content instead of post_content_filtered #25
- [x] Snapshot UUID should be stripped from the URL and JS state once published #37 
- [x] Add UI for listing and accessing current snapshots #12
- [x] Eliminate storage of non-dirty settings #42 
- [x] Suspend kses filters when restoring a revision
- [x] Restore coverage to at least 83%+
- [x] Update readme
#### Post Merge
- [x] Deploy to .org
- [x] Tag release


valendesigns:
@westonruter Is there anything else you can think of that we need to get into this release?

valendesigns:
It looks like there may be parts of https://github.com/xwp/wp-customize-snapshots/commit/ac9709a2dcc4cf50d06e515d35bd667779e4ba15 that could be added?

westonruter:
@valendesigns I think it's good to go

Issue #15, opened 2016-04-26
Allow a snapshot to be scheduled


westonruter:
This would implement https://core.trac.wordpress.org/ticket/28721

A snapshot should be scheduleable from the the Customizer via something like a “Schedule” button along with a timestamp for when it should apply.

If there is also an edit post screen for viewing a snapshot in the admin ( #18), the snapshot should also be able to be scheduled using the traditional post scheduling functionality.

westonruter:
See also the ability to split up a snapshot into multiple: #16

westonruter:
@valendesigns here is the code from the transactions PR for handling the publishing of a transaction post: https://github.com/xwp/wordpress-develop/blob/97fd5019c488a0713d34b517bdbff67c62c48a5d/src/wp-includes/theme.php#L2071-L2120

For snapshots I suppose this logic would have to be short-circuited if `customize_save` is happening already. It would be only done if the snapshot post is being published by the user in the WP admin or if WP Cron is transitioning the future-snapshot to publish. See https://github.com/xwp/wordpress-develop/blob/97fd5019c488a0713d34b517bdbff67c62c48a5d/src/wp-includes/class-wp-customize-transaction.php#L366-L372

westonruter:
There may need to be a “Snapshots” Customizer section that contains the fields for the future published date. This section could also contain a UI for selecting another snapshot to load, and also to provide a name for the snapshot (commit message), as in #19.

See also the (defunct) [Settings Revisions](https://wordpress.org/plugins/settings-revisions/) plugin:

![image](https://cloud.githubusercontent.com/assets/134745/15984102/84b43e08-2f73-11e6-8c9f-c3ec36d62596.png)

Nevertheless, I think it would be ideal if the controls for managing the snapshot remained in the header bar since they should be more meta.

valendesigns:
The issue I see with using the header bar is that switching could be problematic in the small amount of real estate we have up there. I guess we could have an edit button that opens a dialog which allows naming and switching? Or using a key event like control+click or similar.

westonruter:
Note regarding calling `$setting->save()` during publishing: since the snapshot could have settings added to it from multiple users who don't all have the capability to modify each other's settings, or especially in the case of WP Cron transitioning a snapshot's status to `publish`, it will be important to ensure that during the routing for loading the snapshot to save all settings, that a filter be added for `user_has_cap` to grant everything so that every setting will be able to be saved.

westonruter:
[...]

PR #38, opened 2016-06-09
Improve WP Admin UI

westonruter:


php/class-customize-snapshot-manager.php
```diff
@@ -284,7 +289,7 @@ public function create_post_type() {
 	 */
 	function setup_metaboxes() {
 		$id = self::POST_TYPE;
-		$title = __( 'Data', 'customize-snapshots' );
+		$title = __( 'Data', 'customize-snapshot' );
```
valendesigns: The text domain is `customize-snapshots`

php/class-plugin.php
```diff
@@ -74,13 +74,23 @@ public function register_styles( \WP_Styles $wp_styles ) {
...
+	 * @param array $caps    Actual capabilities for meta capability.
 	 * @return array All caps.
 	 */
-	public function filter_user_has_cap( $allcaps ) {
+	public function filter_user_has_cap( $allcaps, $caps ) {
 		if ( ! empty( $allcaps['edit_theme_options'] ) ) {
 			$allcaps['customize_publish'] = true;
 		}
 
+		// Grant all customize snapshot caps which weren't explicitly disallowed to users who can customize.
```
valendesigns, line 85: This is causing test failures.

php/class-customize-snapshot-manager.php
```diff
@@ -494,9 +502,6 @@ public function update_snapshot() {
 		if ( ! check_ajax_referer( self::AJAX_ACTION, 'nonce', false ) ) {
 			status_header( 400 );
 			wp_send_json_error( 'bad_nonce' );
-		} elseif ( ! current_user_can( 'customize' ) ) {
```
valendesigns, line 498: Actually this is what is causing the test failures. Not sure why you want to remove this?

westonruter, line 498: @valendesigns because I added it below instead. The `create_posts` cap is now mapped to `do_not_allow` in the post type so as to hide the “Add New” links for snapshots in the admin. So then we have to use `customize` cap for this instead directly here, which would be doing below anyway.

valendesigns, line 498: You will never get pass this condition `} elseif ( empty( $this->unsanitized_snapshot_post_data ) ) {` because `current_user_can( 'customize' )` will be false in `Customize_Snapshot_Manager::capture_unsanitized_snapshot_post_data`

westonruter, line 498: I restored the `customize` cap check.

php/class-plugin.php
westonruter, line 85: I restored the initial cap check in ab6ca42. Tests are passing now.

php/class-customize-snapshot-manager.php
```diff
@@ -348,6 +348,16 @@ function render_data_metabox( $post ) {
...
 
+		$args = array(
+			'customize_snapshot_uuid' => $post->post_name,
+		);
+		$customize_url = add_query_arg( array_map( 'rawurlencode', $args ), wp_customize_url() );
+		echo sprintf(
+			'<p><a href="%s" class="button button-secondary">%s</a></p>',
+			esc_url( $customize_url ),
+			esc_html__( 'Open in Customizer', 'customize-snapshots' )
+		);
```
valendesigns: This commit makes me happy 😄

valendesigns:
@westonruter When you publish the snapshot all the settings disappear from the metabox

valendesigns:
Oh they are hidden after published. So we need to remove the checkbox toggle and hidden attribute for published snapshots so all the settings display

westonruter:
@valendesigns I think the problem is that all of the settings get sent with `dirty: false` at the moment of Publish. We should preserve the original `dirty` state.

westonruter:
@valendesigns ok, ready for you to review.

valendesigns:
@westonruter I think we should hide the "Save Draft" button or change the customize button to "Edit" so it fits in the space. Thoughts?

valendesigns:
We also need to handle the publish button so the Snapshot updates the front-end. Otherwise scheduling will not work.

valendesigns:
@westonruter An issue with the updated state replacement JS is that the `?` is being left behind when there are no params in the url. It should check that there are params and remove the question mark if there aren't any.

westonruter:
@valendesigns 

> I think we should hide the "Save Draft" button or change the customize button to "Edit" so it fits in the space. Thoughts? We also need to handle the publish button so the Snapshot updates the front-end. Otherwise scheduling will not work.

Actually, perhaps we should just hide the publish box entirely. This would be what #15 would implement, and at that time, the publish button could be restored.

westonruter:
I think I know what to do. We can just remove the `publish_posts` cap for now and it will just work 😄

westonruter:
@valendesigns oh, and you know what? Now with b3ea60a it would be sweet if when a user lacks `customize_publish` capability, in addition to a Save button there could be a “Submit” button that serves the same purpose as the “Submit for Review” button when a user wants to submit a post as pending status:

image

valendesigns:
So it makes the publish button say submit for review?

westonruter:
@valendesigns yeah, try this: https://github.com/xwp/wp-customize-snapshots/compare/feature/wp-admin...feature/submit-for-review

westonruter:
@valendesigns I merged `feature/submit-for-review` into this branch. I'm feeling really good about it now. I think I addressed all of your concerns.

PR #46, opened 2016-06-11
Show list of settings in snapshot in excerpt

### LINKED ISSUE #42: Eliminate storage of non-dirty settings
I think I know why I was having environment problems yesterday. I think it was because I had so many settings loaded into the Customizer (278) and many of the settings were large being post settings with potentially very large `post_content`. I just looked at a snapshot I made and I calculated its size at 76,170 bytes (via `wp post get 305 --field=post_content | wc -c`). This is _large_. The `post_content` field is a `LONGTEXT` in MySQL which means it can store a snapshot `post_content` that is 56,386 times larger. Nevertheless, this is a ton of data to be passing back and forth with each snapshot update, and it could get is into trouble with timeouts and network latency.

I think we should perhaps remove the `full` scope from being a feature for the sake of scalability. With the UX changes in #30 it's not even an option anymore in the UI, and I think that it has always been a bit dubious in its usefulness. Thoughts?

### LINKED ISSUE #45: Filter snapshot excerpt to be list of (dirty) setting IDs
This would be useful for giving a preview for what is contained in the snapshot from the list table:

snapshots_ _wordpress_develop_ _wordpress

See also #42.

westonruter:
image

Fixes #45.
Fixes #42.

### DIFF
--- a/js/customize-snapshots.js
+++ b/js/customize-snapshots.js
@@ -136,10 +136,11 @@
 
 			if ( component.data.isPreview ) {
 				api.each( function( value, key ) {
-					allCustomized[ key ] = {
-						'value': value(),
-						'dirty': value._dirty
-					};
+					if ( value._dirty ) {
+						allCustomized[ key ] = {
+							'value': value()
+						};
+					}
 				} );
 				retval.snapshot_customized = JSON.stringify( allCustomized );
 				retval.snapshot_uuid = component.data.uuid;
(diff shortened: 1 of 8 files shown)


westonruter:
@valendesigns I found lots of references to `dirty` even though it now was basically unused, in 31602e0.

valendesigns:
The reference to dirty is to add only settings to the snapshot that are dirty. It doesn't have anything to do with dirty or full scope during the playback.

westonruter:
We shouldn't be adding things to a snapshot if they aren't dirty to begin with though.

valendesigns:
That's what the dirty check in set was for. 😀

westonruter:
@valendesigns I mean from the JS side of things. We shouldn't be sending a request that contains both dirty and non-dirty settings, right?

westonruter:
@valendesigns there, see deb0ddd.

valendesigns:
Yes, that looks right now.

PR #50, opened 2016-06-15
[WIP] Support Customize Menus

### LINKED ISSUE #2: New Menus are not being saved to the correct location when you publish a snapshot
The menu is being created when the snapshots is published. However, `Menu locations` is not getting checked or updated in the dropdown. Not sure if we need to handle this with JS using something like `wp.customize.bind( 'saved', function () { ... } )` or there is an issue with the data being passed to `customized`.

valendesigns:
Fixes #2

### DIFF
--- a/php/class-customize-snapshot-manager.php
+++ b/php/class-customize-snapshot-manager.php
@@ -124,6 +124,8 @@ class Customize_Snapshot_Manager {
 		add_action( 'admin_bar_menu', array( $this, 'customize_menu' ), 41 );
 		add_action( 'customize_controls_print_footer_scripts', array( $this, 'render_templates' ) );
 
+		add_filter( 'theme_mod_nav_menu_locations', array( $this, 'filter_theme_mod_nav_menu_locations' ) );
+
 		/*
 		 * Add WP_Customize_Widget component hooks which were short-circuited in 4.5 (r36611 for #35895).
 		 * See https://core.trac.wordpress.org/ticket/35895
@@ -378,6 +380,25 @@ class Customize_Snapshot_Manager {
 		add_action( 'wp_restore_post_revision', function() use ( $that ) {
 			$that->restore_kses();
 		} );
+	}
+
+	/**
+	 * Filter for displaying the Snapshot menu location values.
+	 *
+	 * @param array $menu_locations Default menu locations.
+	 * @return array Modified menu locations.
+	 */
+	public function filter_theme_mod_nav_menu_locations( $menu_locations ) {
+		if( isset( $_GET['customize_snapshot_uuid'] ) ) {
+			$mods = $this->snapshot->values();
+			$locations = get_registered_nav_menus();
+			foreach( $locations as $location => $name ) {
+				if( isset( $mods['nav_menu_locations[' . $location . ']'] ) ) {
+					$menu_locations[$location] = $mods['nav_menu_locations[' . $location . ']'];
+				}
+			}
+		}
+		return $menu_locations;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


php/class-customize-snapshot-manager.php
```diff
@@ -123,6 +123,11 @@ public function __construct( Plugin $plugin ) {
 		add_action( 'customize_save_after', array( $this, 'save_snapshot' ) );
 		add_action( 'admin_bar_menu', array( $this, 'customize_menu' ), 41 );
 		add_action( 'customize_controls_print_footer_scripts', array( $this, 'render_templates' ) );
+		
+		add_filter( 'theme_mod_nav_menu_locations', array( $this, 'filter_theme_mod_nav_menu_locations' ) );
+		add_filter( 'wp_get_nav_menus', array( $this, 'filter_wp_get_nav_menus' ) );
+		add_filter( 'wp_get_nav_menu_items', array( $this, 'filter_wp_get_nav_menu_items' ), 10, 3 );
+		add_filter( 'wp_get_nav_menu_object', array( $this, 'filter_wp_get_nav_menu_object' ), 10, 2 );
```
westonruter, line 130: Would we not be able to re-use the methods in `WP_Customize_Nav_Menus`? https://github.com/xwp/wordpress-develop/blob/57d2c46591e6593c8d6bfcb2100b6538e80f9dce/src/wp-includes/customize/class-wp-customize-nav-menu-setting.php#L231-L234

That is, use a similar approach to re-using methods from `WP_Customize_Widgets`?

miina, line 130: Actually while debugging I saw that these filters are already used via `Customize_Snapshot_Manager::preview()`.
https://github.com/xwp/wp-customize-snapshots/blob/master/php/class-customize-snapshot-manager.php#L862-L887

If the `$setting` is a menu setting, the filters are set by `WP_Customize_Nav_Menu_Setting::preview()` but they are not making a difference right now for the Menus section. They do work in case of widgets, e.g. the menus created via snapshot are displayed in the selection list when adding a `Custom Menu` widget.

I'm right now testing the filters used in `WP_Customize_Nav_Menus` as suggested to see if there's still something missing.

miina, line 130: Update: the issue is that these filters are used only settings-specifically but the menus are initially loaded at this moment in `WP_Customize_Nav_Menus`:
https://github.com/xwp/wordpress-develop/blob/master/src/wp-includes/class-wp-customize-nav-menus.php#L509

Loading the `Customize_Nav_Menu_Setting` by itself to use the filters would require being related to existing setting IDs which are unknown before loading the Snapshot data. Any ideas how to still use these instead of creating new filters? Otherwise it seems that new filters are needed.

There's a second issue as well. The following function doesn't return a correct value when used with negative menu ID-s:
https://github.com/xwp/wordpress-develop/blob/57d2c46591e6593c8d6bfcb2100b6538e80f9dce/src/wp-includes/customize/class-wp-customize-nav-menu-section.php#L37-L42
E.g. `nav_menu[-123]` returns 0.

zzramesses:
I tested an existing menu I had on a site.  I edited the placement/order of items and sub-items in this menu.  I also added items to the menu.  This menu wasn't currently set to a location.  I set the menu to a registered menu location.  All of these actions resulted in the information being saved in the snapshot correctly.  I even went into the snapshot and made edits and saw these changes reflected

valendesigns:
Ok, now we need to add back the lost coverage and test with the REST API.

valendesigns:
Closing in favor of #55 which would merge into `develop`.

PR #51, opened 2016-06-16
Added a toolbar link to edit post screen view.

### LINKED ISSUE #48: Add admin bar link to view snapshot edit post screen
In addition to the Customize link opening the snapshot in the Customizer, there should also be an easy way to access the Snapshot edit post screen from the frontend, if browsing the site in a snapshot state (see #47).

miina:
Right now the link name is `Snapshot in Dashboard`. Better suggestions are welcome, I'm not the best word-smith in the world.

Fixes #48.


westonruter:
I changed it to “View Snapshot” and let it be a submenu below “Customize”:

image

PR #54, opened 2016-07-01
Check for unauthorized, unrecognized, and invalid settings prior to sanitizing

westonruter:
Improves logic and alignment with change coming in [#37247](https://core.trac.wordpress.org/ticket/37247): Sanitization needs to come after validation in WP_Customize_Manager


westonruter:
Committed to `trunk`: https://core.trac.wordpress.org/changeset/37942

PR #55, opened 2016-07-02
Support Customize Menus

### LINKED ISSUE #2: New Menus are not being saved to the correct location when you publish a snapshot

valendesigns:
Fixes #2


valendesigns:
@miina & @westonruter I'm thinking we may need to handle the transition to publish in this PR before it's ready for merge. Thought?

westonruter:
@valendesigns transition to publish in what way?

tests/php/test-class-customize-snapshot-manager.php
```diff
@@ -171,6 +184,183 @@ function test_create_post_type() {
...
+		wp_set_current_user( $this->user_id );
+		$this->do_customize_boot_actions( true );
+
+		$manager = new Customize_Snapshot_Manager( $this->plugin );
+		$manager->set_snapshot_uuid();
+		$manager->save_snapshot();
+		$manager->snapshot()->is_preview = true;
+
+		$menu_items = array();
+		$menu_items[] = $this->manager->value_as_wp_post_nav_menu_item(
```
westonruter: I think we can avoid duplicating the `WP_Customize_Nav_Menu_Item_Setting::value_as_wp_post_nav_menu_item()` method here by doing:

``` php
$setting_id = sprintf( 'nav_menu_item[%d]', static::MENU_ID );
$nav_menu_item = new WP_Customize_Nav_Menu_Item_Setting( $this->customize_manager, $setting_id, array(
    value: array( /* ... */ )
) );
$menu_items[] = $nav_menu_item-> value_as_wp_post_nav_menu_item();
```

php/class-customize-snapshot-manager.php
```diff
@@ -381,6 +389,235 @@ function suspend_kses_for_snapshot_revision_restore() {
...
+					unset( $items[ $key ] );
+				}
+			}
+		}
+
+		foreach ( $values as $setting_id => $item ) {
+			if ( preg_match( \WP_Customize_Nav_Menu_Item_Setting::ID_PATTERN, $setting_id, $matches ) ) {
+				if ( (int) $menu->term_id === (int) $item['nav_menu_term_id'] ) {
+					$item['post_id'] = intval( $matches['id'] );
+					$items[] = $this->value_as_wp_post_nav_menu_item( (object) $item );
```
westonruter: I believe we can re-use `WP_Customize_Nav_Menu_Item_Setting::value_as_wp_post_nav_menu_item()` here.

``` php
if ( preg_match( \WP_Customize_Nav_Menu_Item_Setting::ID_PATTERN, $setting_id, $matches ) ) {
    $wp_customize->register_dynamic_settings( array( $setting_id ) );
    $nav_menu_item_setting = $wp_customize->get_setting( $setting_id );
    if ( $nav_menu_item_setting && (int) $menu->term_id === (int) $item['nav_menu_term_id'] ) {
        $item['post_id'] = intval( $matches['id'] );
        $items[] = $nav_menu_item_setting->value_as_wp_post_nav_menu_item(); // <========
    }
}
```

php/class-customize-snapshot-manager.php
```diff
@@ -381,6 +389,235 @@ function suspend_kses_for_snapshot_revision_restore() {
...
+	 * Filter wp_get_nav_menu_items() to load Snapshot values.
+	 *
+	 * @see wp_get_nav_menu_items()
+	 *
+	 * @param array  $items An array of menu item post objects.
+	 * @param object $menu  The menu object.
+	 * @param array  $args  An array of arguments used to retrieve menu item objects.
+	 * @return array Array of menu items.
+	 */
+	function filter_wp_get_nav_menu_items( $items, $menu, $args ) {
```
westonruter: I wonder why this is needed as opposed to `WP_Customize_Nav_Menu_Item::preview()` which adds its own filter handler of the same name? https://github.com/xwp/wordpress-develop/blob/266ffe1f35e577954817d453cd7b2d84778cb57c/src/wp-includes/customize/class-wp-customize-nav-menu-item-setting.php#L368-L480

php/class-customize-snapshot-manager.php
```diff
@@ -381,6 +389,235 @@ function suspend_kses_for_snapshot_revision_restore() {
...
+
+	/**
+	 * Filter wp_get_nav_menus() to load Snapshot values.
+	 *
+	 * @see wp_get_nav_menus()
+	 *
+	 * @param array $menus Array of menus.
+	 * @return array Modified array of menus.
+	 */
+	public function filter_wp_get_nav_menus( $menus ) {
```
westonruter: I wonder why this is needed as opposed to what `WP_Customize_Nav_Menu_Setting::preview()` adds? https://github.com/xwp/wordpress-develop/blob/master/src/wp-includes/customize/class-wp-customize-nav-menu-setting.php#L204-L296

php/class-customize-snapshot-manager.php
miina: The issue is that these filters are used only settings-specifically but the menus are initially loaded at this moment in `WP_Customize_Nav_Menus`:
https://github.com/xwp/wordpress-develop/blob/master/src/wp-includes/class-wp-customize-nav-menus.php#L510

Loading the `Customize_Nav_Menu_Setting` itself to use the filters would require being related to existing setting IDs which are unknown before loading the Snapshot data.

westonruter:
@valendesigns @miina I believe this will reduce the overall amount of code and re-use what is in `WP_Customzie_Nav_Menus` already: https://github.com/xwp/wp-customize-snapshots/pull/56

With those changes merged into this PR, the resulting diff is much smaller: https://github.com/xwp/wp-customize-snapshots/compare/develop...bugfix/saving-menus-issue_3

I tried creating a new nav menu, assigning it to a nav menu location, and adding new nav menu items to it. I created the snapshot and loaded it on the frontend in both an authenticated and incognito window, and the snapshotted menu and nav menu items appeared as expected for me in both cases.

Please review.

westonruter:
I think this is finally ready! 😂

PR #62, opened 2016-07-20
Allow snapshot posts to publish and be scheduled

### LINKED ISSUE #15: Allow a snapshot to be scheduled

PatelUtkarsh:
This allows snapshot changes to be published, On publish it will save snapshot changes in database.

Fixes #15.

### DIFF
--- a/php/class-post-type.php
+++ b/php/class-post-type.php
@@ -39,6 +39,12 @@ class Post_Type {
 	protected $kses_suspended = false;
 
 	/**
+	 * Surpress hook call
+	 *
+	 * @var bool
+	 */
+	static protected $suppress_publish_hook = false;
+	/**
 	 * Constructor.
 	 *
 	 * @access public
@@ -567,7 +573,7 @@ class Post_Type {
 	 * @param \WP_Post $post Post object.
 	 */
 	public function publish_snapshot( $post_id, $post ) {
-		if ( did_action( 'customize_save' ) ) {
+		if ( did_action( 'customize_save' ) || static::$suppress_publish_hook ) {
 			// Short circuit because customize_save ajax call is changing status.
 			return;
 		}
@@ -645,19 +651,12 @@ class Post_Type {
 		}
 
 		if ( true === $have_error ) {
-			$action = 'publish_' . static::SLUG;
-			$callback = array( $this, 'publish_snapshot' );
-			$priority = has_action( $action, $callback );
-			if ( false !== $priority ) {
-				remove_action( $action, $callback, $priority );
-			}
+			static::$suppress_publish_hook = true;
 			wp_update_post( array(
 				'ID' => $post->ID,
 				'post_content' => Customize_Snapshot_Manager::encode_json( $snapshot_content ),
 			) );
-			if ( false !== $priority ) {
-				add_action( $action, $callback, $priority, 2 );
-			}
+			static::$suppress_publish_hook = false;
 		} else {
 			// Remove any previous error on setting.
 			delete_post_meta( $post_id, 'snapshot_error_on_publish' );
@@ -674,8 +673,7 @@ class Post_Type {
 		if ( false !== $priority ) {
 			remove_action( $action, $callback, $priority );
 		}
-		/** This action is documented in wp-includes/class-wp-customize-manager.php */
-		do_action( 'customize_save', $this->snapshot_manager->customize_manager );
+		$this->snapshot_manager->customize_manager->doing_ajax( 'customize_save' );
 		if ( false !== $priority ) {
 			add_action( $action, $callback, $priority, 0 );
 		}
--- a/php/class-post-type.php
+++ b/php/class-post-type.php
@@ -635,7 +635,8 @@ class Post_Type {
 		<style>
 			#misc-publishing-actions .misc-pub-post-status,
 			#misc-publishing-actions .misc-pub-visibility,
-			#misc-publishing-actions .misc-pub-curtime {
+			#misc-publishing-actions .misc-pub-curtime,
+			.submitbox #publish {
 				display: none;
 			}
 		</style>
--- a/php/class-post-type.php
+++ b/php/class-post-type.php
@@ -108,6 +108,7 @@ class Post_Type {
 		add_filter( 'wp_insert_post_data', array( $this, 'preserve_post_name_in_insert_data' ), 10, 2 );
 		add_filter( 'user_has_cap', array( $this, 'filter_user_has_cap' ), 10, 2 );
 		add_action( 'publish_' . static::SLUG, array( $this, 'publish_snapshot' ), 10, 2 );
+		add_filter( 'display_post_states', array( $this, 'display_post_states' ), 10, 2 );
 	}
 
 	/**
@@ -346,14 +347,20 @@ class Post_Type {
 		ksort( $snapshot_content );
 		echo '<ul id="snapshot-settings">';
 		foreach ( $snapshot_content as $setting_id => $setting_params ) {
-			if ( ! isset( $setting_params['value'] ) ) {
+			if ( ! isset( $setting_params['value'] ) && ! isset( $setting_params['save_error'] ) ) {
 				continue;
 			}
-			$value = $setting_params['value'];
-
+			$value = isset( $setting_params['value'] ) ? $setting_params['value'] : '';
 			echo '<li>';
 			echo '<details open>';
-			echo '<summary><code>' . esc_html( $setting_id ) . '</code></summary>';
+			echo '<summary><code>' . esc_html( $setting_id ) . '</code>';
+			if ( isset( $setting_params['save_error'] ) ) {
+				echo '<span class="attention">';
+				echo '<b>' . __( 'Error:', 'customize-snapshots' ) . ' </b>';
+				echo '<code>' . esc_html( $setting_params['save_error'] ) . '</code>';
+				echo '</span>';
+			}
+			echo '</summary>';
 			if ( is_string( $value ) || is_numeric( $value ) ) {
 				$preview = '<p>' . esc_html( $value ) . '</p>';
 			} elseif ( is_bool( $value ) ) {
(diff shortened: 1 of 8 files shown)


PatelUtkarsh:
Do we need to check for theme change and change theme before publishing snapshot data and revert back after done? @valendesigns

php/class-post-type.php
```diff
@@ -549,4 +548,37 @@ public function filter_user_has_cap( $allcaps, $caps ) {
...
 		return $allcaps;
 	}
+
+	/**
+	 * Publish snapshot changes when snapshot post is being published.
+	 *
+	 * @param int      $post_id Post id.
+	 * @param \WP_Post $post Post object.
+	 */
+	public function publish_snapshot( $post_id, $post ) {
```
westonruter: @PatelUtkarsh warning that when doing Publish in the Customizer, any existing snapshot will be transitioned to `publish`. See: https://github.com/xwp/wp-customize-snapshots/blob/f69b32592021cef7ffa5d34419f67f70950f7ec3/php/class-customize-snapshot-manager.php#L591

So it is important that this should short-circuit if `did_action( 'customize_save' )`, because this would indicate that the `customize_save` Ajax request is being made.

php/class-post-type.php
```diff
@@ -549,4 +548,37 @@ public function filter_user_has_cap( $allcaps, $caps ) {
...
+				return ! is_null( $value );
+			}
+		);
+
+		/** This action is documented in wp-includes/class-wp-customize-manager.php */
+		do_action( 'customize_save', $this->snapshot_manager->customize_manager );
+
+		foreach ( $snapshot_values as $setting_id => $setting_params ) {
+			$this->snapshot_manager->customize_manager->set_post_value( $setting_id, $setting_params );
+			$this->snapshot_manager->customize_manager->add_setting( $setting_id, $setting_params );
```
westonruter: @PatelUtkarsh this isn't going to work as expected. Instead, you should above do:

``` php
$this->snapshot_manager->customize_manager->add_dynamic_settings( array_keys( $snapshot_values ) );
```

This will ensure that the dynamic settings are added. For static settings, it will be important to do:

``` php
do_action( 'customize_register', $this->snapshot_manager->customize_manager );
```

php/class-post-type.php
```diff
@@ -549,4 +548,37 @@ public function filter_user_has_cap( $allcaps, $caps ) {
...
+	 *
+	 * @param int      $post_id Post id.
+	 * @param \WP_Post $post Post object.
+	 */
+	public function publish_snapshot( $post_id, $post ) {
+		if ( ! is_a( $this->snapshot_manager->customize_manager, '\WP_Customize_Manager' ) ) {
+			require_once( ABSPATH . WPINC . '/class-wp-customize-manager.php' );
+			$this->snapshot_manager->customize_manager = new \WP_Customize_Manager();
+		}
+		$snapshot_content = json_decode( $post->post_content, true );
```
westonruter: Instead of accessing the `post_content` directly, it would be better to do:

``` php
$snapshot_content = $this->snapshot_manage->post_type->get_post_content( $post );
```

westonruter:
@PatelUtkarsh:

> Do we need to check for theme change and change theme before publishing snapshot data and revert back after done?

 I don't think we need to worry about the theme change. The publish metabox and publish actions should be disabled if the current theme doesn't match the theme associated with the snapshot.

westonruter:
@PatelUtkarsh the changes here will directly resolve #15. Please refer to the comments there for what other considerations will be needed.

php/class-post-type.php
```diff
@@ -549,4 +548,37 @@ public function filter_user_has_cap( $allcaps, $caps ) {
...
+			}
+		);
+
+		/** This action is documented in wp-includes/class-wp-customize-manager.php */
+		do_action( 'customize_save', $this->snapshot_manager->customize_manager );
+
+		foreach ( $snapshot_values as $setting_id => $setting_params ) {
+			$this->snapshot_manager->customize_manager->set_post_value( $setting_id, $setting_params );
+			$this->snapshot_manager->customize_manager->add_setting( $setting_id, $setting_params );
+			$setting = $this->snapshot_manager->customize_manager->get_setting( $setting_id );
```
westonruter: @PatelUtkarsh you'll need to make sure that `$setting` is not `null`. Also, as noted in #15, you'll need to make sure you do: `$setting->capability = 'exist'` to ensure that the setting save will always be allowed. This is safe because a setting only be written to the snapshot by a user who has the capability to begin with, so we know at this point it will be safe to save.

php/class-post-type.php
```diff
@@ -549,4 +548,37 @@ public function filter_user_has_cap( $allcaps, $caps ) {
...
+			$setting->save();
+		}
```
westonruter, line 662: @PatelUtkarsh make sure you also `do_action( 'customize_save_after', $this->snapshot_manager->customize_manager )`

php/class-post-type.php
```diff
@@ -549,4 +548,37 @@ public function filter_user_has_cap( $allcaps, $caps ) {
...
+	 * Publish snapshot changes when snapshot post is being published.
+	 *
+	 * @param int      $post_id Post id.
+	 * @param \WP_Post $post Post object.
+	 */
+	public function publish_snapshot( $post_id, $post ) {
+		if ( ! is_a( $this->snapshot_manager->customize_manager, '\WP_Customize_Manager' ) ) {
+			require_once( ABSPATH . WPINC . '/class-wp-customize-manager.php' );
+			$this->snapshot_manager->customize_manager = new \WP_Customize_Manager();
+		}
```
westonruter, line 620: Add this:

``` php
if ( ! did_action( 'customize_register' ) ) {
    do_action( 'customize_register', $this->snapshot_manager->customize_manager );
}
```

php/class-post-type.php
```diff
@@ -549,4 +548,37 @@ public function filter_user_has_cap( $allcaps, $caps ) {
...
+		$snapshot_content = json_decode( $post->post_content, true );
+		if ( json_last_error() || ! is_array( $snapshot_content ) ) {
+			return;
+		}
+		$snapshot_values = array_filter(
+			wp_list_pluck( $snapshot_content, 'value' ),
+			function( $value ) {
+				return ! is_null( $value );
+			}
+		);
```
westonruter: You'll also want to do this here, to ensure all of the expected actions are done as found in `WP_Customize_Manager::save()`:

``` php
if ( method_exists( $this->snapshot_manager->customize_manager, 'validate_setting_values' ) ) {
    do_action( 'customize_save_validation_before', $this->snapshot_manager->customize_manager );
}
```

php/class-post-type.php
```diff
@@ -154,7 +154,6 @@ public function setup_metaboxes() {
 	 */
 	public function remove_publish_metabox() {
 		remove_meta_box( 'slugdiv', static::SLUG, 'normal' );
-		remove_meta_box( 'submitdiv', static::SLUG, 'side' );
```
westonruter, line 179: Keep removing if the snapshot's associated theme is not equal to the current theme.

PatelUtkarsh:
@westonruter Which post_status transition should we allow? 

```
any => publish or 
only draft => publish
```

westonruter:
@PatelUtkarsh any transition to `publish`. Particularly, this is key for scheduling snapshots (#15), as it will allow you to create update a snapshot post with a future date and it will have `future` status. When the future date arrives, WP Cron should transition the post from `future` to `publish` and all of the Customizer settings in that post should get saved. (If the theme associated with the snapshot doesn't match the current theme, then we should probably short-circuit and change the status back to `pending`.)

php/class-post-type.php
```diff
@@ -549,4 +552,86 @@ public function filter_user_has_cap( $allcaps, $caps ) {
...
+		}
+
+		foreach ( $filtered_setting_objs as $setting_obj ) {
+			$setting_obj->capability = 'exist';
+		}
+
+		remove_action( 'customize_save', array( $this->snapshot_manager, 'check_customize_publish_authorization' ), 10 );
+		/** This action is documented in wp-includes/class-wp-customize-manager.php */
+		do_action( 'customize_save', $this->snapshot_manager->customize_manager );
+		add_action( 'customize_save', array( $this, 'check_customize_publish_authorization' ), 10, 0 );
```
westonruter: Looks like the wrong `customize_save` handler is being re-added. There is no `check_customize_publish_authorization` method on this class.

I suggest doing:

``` php
$action = 'customize_save';
$callback = array( $this->snapshot_manager, 'check_customize_publish_authorization' );
$priority = has_action( $action, $callback );
if ( false !== $priority ) {
    remove_action( $action, $callback, $priority );
}
/** This action is documented in wp-includes/class-wp-customize-manager.php */
do_action( 'customize_save', $this->snapshot_manager->customize_manager );
if ( false !== $priority ) {
    add_action( $action, $callback, $priority );
}
```

php/class-post-type.php
```diff
@@ -549,4 +552,86 @@ public function filter_user_has_cap( $allcaps, $caps ) {
...
+
+		$filtered_setting_objs = array_filter( $setting_objs, function( $setting_obj ) {
+			return is_a( $setting_obj, '\WP_Customize_Setting' );
+		} );
+		if ( empty( $filtered_setting_objs ) ) {
+			return;
+		}
+
+		foreach ( $filtered_setting_objs as $setting_obj ) {
+			$setting_obj->capability = 'exist';
```
westonruter: We should probably capture what the original `capability` was before we override it, and then restore the original `capability` after we `save` them all. This will ensure that a setting doesn't accidentally get leaked out with blanket permissions.

php/class-post-type.php
```diff
@@ -549,4 +552,86 @@ public function filter_user_has_cap( $allcaps, $caps ) {
...
+
+		$this->snapshot_manager->customize_manager->add_dynamic_settings( array_keys( $snapshot_values ) );
+		$setting_objs = array();
+		foreach ( $snapshot_values as $setting_id => $setting_params ) {
+			$this->snapshot_manager->customize_manager->set_post_value( $setting_id, $setting_params );
+			$setting_objs[] = $this->snapshot_manager->customize_manager->get_setting( $setting_id );
+		}
+
+		$filtered_setting_objs = array_filter( $setting_objs, function( $setting_obj ) {
+			return is_a( $setting_obj, '\WP_Customize_Setting' );
```
westonruter: I think it's better to do `return $setting_obj instanceof \WP_Customize_Setting`. It's more modern.

php/class-post-type.php
```diff
@@ -549,4 +552,86 @@ public function filter_user_has_cap( $allcaps, $caps ) {
...
+		} );
+		if ( empty( $filtered_setting_objs ) ) {
+			return;
```
westonruter, line 629: I've been thinking about this, about the scenario where a setting no longer is recognized when a snapshot is published. At the moment, they just silently get skipped. But I think there should be a way to log when a setting fails to save for reasons such as the value no longer being valid or the setting not being recognized. We could introduce a `save_error` perhaps for each setting in a snapshot which gets set when the snapshot is saved, and this could potentially store such failure information.

php/class-post-type.php
```diff
@@ -549,4 +552,86 @@ public function filter_user_has_cap( $allcaps, $caps ) {
...
+	 * @param int      $post_id Post id.
+	 * @param \WP_Post $post Post object.
+	 */
+	public function publish_snapshot( $post_id, $post ) {
+		if ( did_action( 'customize_save' ) ) {
+			// Short circuit because customize_save ajax call is changing status.
+			return;
+		}
+		$snapshot_theme = get_post_meta( get_the_ID(), '_snapshot_theme', true );
+		if ( ! empty( $snapshot_theme ) && get_stylesheet() !== $snapshot_theme ) {
```
westonruter: @valendesigns What were the reasons for preventing snapshots from being published when doing a theme switch? I can't recall. Is it because themes may define their own settings, and if a different theme is active when a snapshot is loaded or saved, it could result in there being unrecognized settings?

westonruter:
@PatelUtkarsh this commit may be especially relevant to you, specifically the `date_gmt` field: https://github.com/xwp/wp-customize-snapshots/pull/63/commits/13f8edfd8813c11b8c3f6f1771d2252a11d3cc63

See https://github.com/xwp/wp-customize-snapshots/pull/63

php/class-post-type.php
```diff
@@ -549,4 +579,147 @@ public function filter_user_has_cap( $allcaps, $caps ) {
...
+			$setting_obj->capability = 'exist';
+		}
+
+		$action = 'customize_save';
+		$callback = array( $this->snapshot_manager, 'check_customize_publish_authorization' );
+		$priority = has_action( $action, $callback );
+		if ( false !== $priority ) {
+			remove_action( $action, $callback, $priority );
+		}
+		$this->snapshot_manager->customize_manager->doing_ajax( 'customize_save' );
```
westonruter: The `doing_ajax` method doesn't do anything other than return a `bool` to say whether this Ajax action is happening. So I don't think this block is doing anything. I'm revising.

PatelUtkarsh:
@westonruter In a case of null as setting value before publishing snapshot, it is not showing setting name, and after publishing it shows setting name with an error.
Should we fix this and show null setting name before publishing?

PR #68, opened 2016-07-28
Schedule UI for snapshot in customizer

PatelUtkarsh:


js/customize-snapshots.js
```diff
@@ -40,10 +52,20 @@
 
 			component.extendPreviewerQuery();
 			component.addButtons();
+			component.addSlideDown();
 
 			$( '#snapshot-save' ).on( 'click', function( event ) {
+				var scheduleDate;
 				event.preventDefault();
-				component.sendUpdateSnapshotRequest( { status: 'draft' } );
+				if ( $( this ).html() === component.data.i18n.scheduleButton && ! _.isEmpty( component.snapshotScheduleBox ) && component.getDateFromInputs() && component.isScheduleDateFuture() ) {
```
westonruter: `$( this ).html()` seems dirty. Can't you look at `snapshotStatus` instead?

js/customize-snapshots.js
```diff
@@ -151,6 +173,96 @@
 		return a.href;
 	};
 
+	component.dateComponentInputs = {};
+
+	component.snapshotScheduleBox = {};
+
+	component.addSlideDown = function slideDown() {
```
westonruter: Add JSDoc.

js/customize-snapshots.js
```diff
@@ -151,6 +173,96 @@
...
+
+		api.state( 'snapshot-exists' ).bind( function( exists ) {
+			if ( exists && ! _.isEmpty( component.snapshotScheduleBox ) ) {
+				component.updateSnapshotScheduleBox();
+			}
+		} );
+
+	};
+
+	component.updateSnapshotScheduleBox = function updateSnapshotScheduleBox() {
```
westonruter: Add JSDoc.

westonruter:
ESLint errors:

```
js/customize-snapshots.js: line 212, col 54, Error - 'e' is already declared in the upper scope. (no-shadow)
js/customize-snapshots.js: line 261, col 72, Error - 'component' is already declared in the upper scope. (no-shadow)
js/customize-snapshots.js: line 531, col 62, Error - No magic number: 1. (no-magic-numbers)
js/customize-snapshots.js: line 575, col 55, Error - No magic number: 1. (no-magic-numbers)
js/customize-snapshots.js: line 613, col 39, Error - Expected to return a value at the end of this function. (consistent-return)
js/customize-snapshots.js: line 626, col 6, Error - Blocks are nested too deeply (4). (max-depth)
js/customize-snapshots.js: line 633, col 5, Error - Unexpected if as the only statement in an else block. (no-lonely-if)
js/customize-snapshots.js: line 648, col 2, Error - Missing JSDoc return description. (valid-jsdoc)
js/customize-snapshots.js: line 660, col 46, Error - No magic number: 1000. (no-magic-numbers)
js/customize-snapshots.js: line 691, col 46, Error - No magic number: 1000. (no-magic-numbers)
```

And phpunit failure:

```
1) CustomizeSnapshots\Test_Ajax_Customize_Snapshot_Manager::test_ajax_update_snapshot_cap_check with data set #4 ('administrator', array(true, array(null, array(true))))
Failed asserting that Array &0 (
    'success' => true
    'data' => Array &1 (
        'errors' => null
        'setting_validities' => Array &2 (
            'anyonecanedit' => true
        )
        'snapshot_publish_date' => '0000-00-00 00:00:00'
    )
) is identical to Array &0 (
    'success' => true
    'data' => Array &1 (
        'errors' => null
        'setting_validities' => Array &2 (
            'anyonecanedit' => true
        )
    )
).
```

js/customize-snapshots.js
PatelUtkarsh: `snapshotStatus` can be `draft` or `future` we can remove that check as we are checking for the future date on the same condition. And will also add `isSnapshotHasUnsavedChanges` check.

.travis.yml
```diff
@@ -16,6 +16,7 @@ env:
     - WP_VERSION=latest WP_MULTISITE=1
 
 install:
+    - nvm install 4 && nvm use 4
```
valendesigns, line 19: Why do we need node version 4, what is this fixing?

westonruter, line 19: ESLint requires Node 4 now.

westonruter, line 19: There was an ESLint fatal error otherwise since Node<4 doesn't support `let`. ESLint would also warn in the Travis build about the old version of node, before ultimately it would fatal error.

js/customize-snapshots.js
```diff
@@ -290,14 +474,15 @@
...
+			customize_snapshot_uuid: component.data.uuid
+		} );
+
 		data = _.extend(
 			{},
 			api.previewer.query(),
-			{
-				nonce: api.settings.nonce.snapshot,
-				customize_snapshot_uuid: component.data.uuid,
-				status: args.status
```
valendesigns, line 299: @PatelUtkarsh I'm not sure I understand this change. Why remove status?

PatelUtkarsh, line 299:
```
data = _.extend( args, {
            nonce: api.settings.nonce.snapshot,
            customize_snapshot_uuid: component.data.uuid
        } );
```

We are merging/extending args so we don't need status to be merge seperately.

valendesigns, line 299: Why do we need to extend args? Are the args being used in the preview query for something else now?

valendesigns:
@westonruter If you get a chance and have the runway please review this for correct/complete functionality. I'm going to go over it again before it's merged Monday, but just wanted a second set of eyes on it.

valendesigns:
@PatelUtkarsh I found an issue. The edit icon should disappear if the Customizer state is dirty.

PatelUtkarsh:
image
UI issue, because of `position: relative; z-index:10000` on `.select2-container`

PatelUtkarsh:
Should we move `#snapshot-schedule` from `#customize-header-actions` to somewhere else? @valendesigns

js/customize-snapshots.js
PatelUtkarsh, line 299: We can optimize this extending (3 times to 1). Here, I am merging args with options above, option param contains status and publish date (Optional: in the case of schedule snapshot) so because we don't need to explicitly check for publish date's existence, I am merging it directly.

valendesigns:
@PatelUtkarsh If you could add a couple more tests to bring the coverage back up then we can merge.

php/class-customize-snapshot-manager.php
```diff
@@ -1322,15 +1358,97 @@ public function remove_all_non_snapshot_admin_bar_links( $wp_admin_bar ) {
...
+				<div class="snapshot-schedule-title">
+					<h3>
+						<?php esc_html_e( 'Schedule Snapshot', 'customize-snapshots' ); ?>
+						<span class="reset-time">(<a href="#"><?php esc_html_e( 'Reset', 'customize-snapshots' ) ?></a>)</span>
+					</h3>
+					<span class="snapshot-schedule-description">
+						<span class="snapshot-scheduled-countdown"></span>
+						<span class="timezone-info"><?php echo esc_html( $date_control_description ); ?></span>
+					</span>
+					<a href="{{ data.editLink }}" class="dashicons dashicons-edit snapshot-edit-link" aria-expanded="false"></a>
```
westonruter: Can this have `target="_blank"`? Also, what about `screen-reader-text` and a `title` tooltip. I don't think it's clear what this edit link is for. In fact, it might even be better moved to the top row since it's really not even specific to scheduling anymore, right?

customize__wordpress_develop_ _just_another_customizer-driven_site

js/customize-snapshots.js
```diff
@@ -399,6 +588,207 @@
...
+	component.isFutureDate = function isFutureDate() {
+		var date = component.getDateFromInputs(),
+			millisecondsDivider = 1000,
+			remainingTime;
+
+		if ( ! date ) {
+			return false;
+		}
+
+		remainingTime = component.dateValueOf( date );
```
westonruter, line 748: Somewhat of a problem here. `date` is a `Date` object. However, the `component.dateValueOf()` function takes a `string` as its param. It looks like `dateValueOf` needs to take a `Date` as its argument as well?

``` diff
diff --git a/js/customize-snapshots.js b/js/customize-snapshots.js
index 40cf91e..6ecd668 100644
--- a/js/customize-snapshots.js
+++ b/js/customize-snapshots.js
@@ -774,19 +774,21 @@
    /**
     * Get the primitive value of a Date object.
     *
-    * @param {string} dateString The post status for the snapshot.
+    * @param {Date|string} [date] The post status for the snapshot. Defaults to current date.
     * @returns {object|string} The primitive value or date object.
     */
-   component.dateValueOf = function( dateString ) {
-       var date;
+   component.dateValueOf = function( date ) {
+       var dateObj;

-       if ( _.isUndefined( dateString ) ) {
-           date = new Date();
+       if ( _.isUndefined( date ) ) {
+           dateObj = new Date();
+       } else if ( 'string' === typeof date ) {
+           dateObj = new Date( date );
        } else {
-           date = new Date( dateString );
+           dateObj = date;
        }

-       return date.valueOf();
+       return dateObj.valueOf();
    };

    component.init();
```

php/class-customize-snapshot-manager.php
PatelUtkarsh: Yes, it's not specific to scheduling. We can move it to the top but we don't have much space left there.
I will add `target` and `screen-reader-text` and `title`.

php/class-customize-snapshot-manager.php
```diff
@@ -634,16 +634,25 @@ public function enqueue_controls_scripts() {
...
+			$post = $this->snapshot->post();
+			$this->override_post_date_default_data( $post );
+		}
 
 		// Script data array.
 		$exports = apply_filters( 'customize_snapshots_export_data', array(
 			'action' => self::AJAX_ACTION,
 			'uuid' => $this->snapshot ? $this->snapshot->uuid() : self::generate_uuid(),
-			'editLink' => $this->snapshot ? get_edit_post_link( $this->snapshot->post(), 'raw' ) : '',
+			'editLink' => $this->snapshot ? get_edit_post_link( $post, 'raw' ) : '',
```
westonruter: Warning: `$post` may not have been defined. Above you should do `$post = null` to initialize, or else we'll get undefined variable warnings.

westonruter:
@PatelUtkarsh something is wrong with the time being entered. I change a date and then click Reset and it always is showing for me as 16:25, no matter what the current time is:

date/time

PatelUtkarsh:
@westonruter Maybe that is the time when snapshot was saved in DB, reset only reset time to snapshot's publish date if publish date is empty it resets to current date/time,

Should we tweak it like this: if snapshot date is future reset to that date or else reset to current date/time?

westonruter:
The `(Reset)` link should be initially shown if the date is not `0000-00-00 00:00:00`? Instead, it is initially hidden:

westonruter:
@PatelUtkarsh ah, thanks for that info. OK, it does make sense to me now I think. The reset link is initially hidden because it behaves differently than it does in Customize Posts. Instead of resetting it to NOW it resets it to the original value. I was expecting _Reset_ to undo the scheduling, by setting the date back to `0000-00-00 00:00:00` so that it would un-schedule and make it go live presently.

js/customize-snapshots.js
PatelUtkarsh, line 748: Fixed `dateValueOf` function.

westonruter:
How in the world did the coverage go DOWN when I added MORE tests?!

PR #70, opened 2016-08-07
Ensure frontend links and Ajax requests include UUID query param; add previewing of Ajax requests and form submissions

### LINKED ISSUE #47: Filter all of generated links to include the snapshot UUID if present in the request
When viewing the frontend of the site, clicking around should preserve the snapshot state as much as possible.

See https://github.com/xwp/wordpress-develop/pull/61/files#diff-4fb8a477f559bdfad2c1e6db6d1c8b06R887

### LINKED ISSUE #65: Inject Customizer state into Ajax requests
This goes hand-in-hand with #47. In order to preview sites that use a lot of JS and Ajax requests, the requests need to be Customzier-aware. We can accomplish this via jQuery Ajax prefilter. If on the frontend, we only need to inject the `customzie_snapshot_uuid` query param. In the Customizer Preview, we'll need to rewrite the requests to be `POST` and ensure the Customzier query vars are included: this has been demonstrated to work for the WP REST API, but it will not work in all cases. See implementation from Customize REST Resources which can be lifted: https://github.com/xwp/wp-customize-rest-resources/blob/84405f4312e22e956e86b6d17f0791d64e8de372/js/rest-resources-manager.js#L208-L265

westonruter:
When previewing a snapshot on the frontend, the admin bar is modified accordingly:

wordpress_develop_food_ _just_another_customizer-driven_site_
- (Unchanged) The “Customize” link takes you to the Customizer with that snapshot state applied.
- The “Inspect Snapshot” link takes you to the edit post screen (formerly submenu item under Customize).
- A new “Exit Snapshot Preview” link appears which removes the `customize_snapshot_uuid` query param from the current URL.
- All other admin bar links are removed, other than those which already contain the `customize_snapshot_uuid` query param (as in the Airplane Mode toggle shown here), or which are in-page links, such as the query monitor link.

When you are navigating around the site with a snapshot applied, the `sessionStorage` will store the UUID for this snapshot. While the JS will attempt to inject the `customize_snapshot_uuid` into all links in the document, if you _do_ navigate away from a URL that has this query param, the JS on the new non-snapshotted page will see that you have a snapshot UUID in `sessionStorage` but no snapshot in the URL, and it will add an admin bar link to prompt you to resume the snapshot preview:

wordpress_develop_ _just_another_customizer-driven_site__and_2__vagrant_vvv_____bash_

This PR also ensures that Ajax requests made on the frontend include the `customize_snapshot_uuid` param so that the responses can be previewed. Additionally, Ajax requests made _inside_ the Customizer preview also now get the Customizer state added so that their responses should be previewable as well. 

Lastly, forms that have `GET` request methods (such as the search form) now are actually usable in the Customizer preview! Likewise, forms on the frontend will persist the current `customize_snapshot_uuid` in the requests. This fixes a long-standing Trac ticket [#20714](https://core.trac.wordpress.org/ticket/20714): Theme customizer: Impossible to preview a search results page. See sub-PR #72.

Fixes #47
Fixes #65

### DIFF
--- a/css/customize-snapshots-preview.css
+++ b/css/customize-snapshots-preview.css
@@ -1,0 +1,8 @@
+form[method="post"] button[type="submit"],
+form[method="post"] button[type=""],
+form[method="post"] input[type="submit"] {
+	cursor: not-allowed;
+}
+form[method="post"] button:not([type]) {
+	cursor: not-allowed;
+}
--- a/customize-snapshots.php
+++ b/customize-snapshots.php
@@ -44,7 +44,7 @@ if ( version_compare( phpversion(), '5.3', '>=' ) ) {
  * Admin notice for incompatible versions of PHP.
  */
 function customize_snapshots_php_version_error() {
-	printf( '<div class="error"><p>%s</p></div>', customize_snapshots_php_version_text() );
+	printf( '<div class="error"><p>%s</p></div>', customize_snapshots_php_version_text() ); // WPCS: XSS OK.
 }
 
 /**
--- a/instance.php
+++ b/instance.php
@@ -23,3 +23,31 @@ function get_plugin_instance() {
 	global $customize_snapshots_plugin;
 	return $customize_snapshots_plugin;
 }
+
+/**
+ * Convenience function for whether settings are being previewed.
+ *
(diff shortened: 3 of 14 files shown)


valendesigns:
@westonruter Looks good! 👍

Issue #60, opened 2016-07-12
Facilitate users concurrently editing a snapshot


westonruter:
If two users edit the same snapshot, there is currently the real possibility that users will override each other's changes. There needs to be concurrency support for snapshots. When this happens, the snapshot save should be rejected as a validation failure, with a prompt to resolve the conflict, by accepting their changes or overriding the changes with their own.

We need to keep track of the settings that have been changed _since the last snapshot save_ and send it along with the snapshot save request. See approach for keeping track of settings changed: https://github.com/xwp/wordpress-develop/blob/a501d6c8be242dce7bcff0e14b145953057ff7ba/src/wp-admin/js/customize-controls.js#L3685-L3688

Whenever a value is updated in a snapshot, beyond storing the setting `value` we can also store the timestamp for when the setting was updated and also the user ID who actually saved it. 

Other general notes on concurrency locking in general:
- Validation is an important way that we can solve concurrency problems.
- Use heartbeat to announce that a setting has changed, and allow the value to be supplied from the other user
- Send locking overrides via separate channel than customized data, since modified date hack won't work for other kinds of settings

lgedeon:
@westonruter Instead of locking could we add a history of changes that have happened in the snapshot and keep the latest value? Then we could show that a change has been made and let the content editors keep, restore a different version or combine?

I am thinking of a ui like this:
<img width="301" alt="screen shot 2016-07-21 at 11 43 02 pm" src="https://cloud.githubusercontent.com/assets/370981/17046066/808cda5c-4f9d-11e6-89bb-1cc466d2434d.png">

Both links (red and yellow) would pop-up a history (new panel?) with the option to go back to an older version.

The code to capture the history could look like this: (modified from /plugins/customize-snapshots/php/class-customize-snapshot.php)

``` php
    public function set( \WP_Customize_Setting $setting, $value, $deprecated = null ) {
        if ( ! is_null( $deprecated ) ) {
            _doing_it_wrong( __METHOD__, 'The $dirty argument has been removed.', '0.4.0' );
            if ( false === $deprecated ) {
                return;
            }
        }

        $history = isset( $this->data[ $setting->id ]['history'] ) ? (array) $this->data[ $setting->id ]['history'] : array();

        if ( ! isset( $this->data[ $setting->id ]['value'] ) || $this->data[ $setting->id ]['value'] !== $value ) {
            $history[] = array(
                'value' => $value,
                'user' => get_current_user_id(),
[...]

PR #71, opened 2016-08-09
Snapshot Conflicts 

### LINKED ISSUE #22: Show conflicts arising with scheduled snapshots
When making a change in the Customizer or making a new snapshot, if there is currently one or more scheduled snapshots that touch on a setting that is dirty in the current state, there should be a warning to note of a potential conflict.

See also previewing future scheduled snapshots: #15 & #21

PatelUtkarsh:
Fixes #22


php/class-post-type.php
```diff
@@ -675,4 +706,55 @@ public function hide_disabled_publishing_actions( $post ) {
...
+	 * Get conflicts settings
+	 *
+	 * @param \WP_Post $post post to compare conflict values.
+	 *
+	 * @return array
+	 */
+	public function get_conflicts_setting( $post ) {
+		global $wpdb;
+		$post = get_post( $post );
+		if ( ! $post || $post instanceof \WP_Error ) {
```
valendesigns: This conditional would probably make more sense like so:

``` PHP
if ( ! $post || ! ( $post instanceof \WP_Post ) ) {
```

php/class-post-type.php
```diff
@@ -675,4 +706,55 @@ public function hide_disabled_publishing_actions( $post ) {
...
+			return array();
+		}
+		$return = array();
+		$content = $this->get_post_content( $post );
+		if ( empty( $content ) || ! is_array( $content ) ) {
+			return $return;
+		}
+		$settings = array_keys( $content );
+		$query = $wpdb->prepare( "SELECT ID, post_name, post_status, post_content FROM $wpdb->posts WHERE post_type = %s AND post_status IN ( 'draft', 'pending', 'future' ) AND ID != %d AND ( 1 = 2 ", self::SLUG, $post->ID );
```
valendesigns: Not sure the `1 = 2` is needed here?

css/customize-snapshots-admin.css
```diff
@@ -0,0 +1,3 @@
+a.snapshot-thickbox{
```
valendesigns: Add a space before curly braces.

php/class-post-type.php
```diff
@@ -380,18 +386,23 @@ public function render_data_metabox( $post ) {
 				}
 				echo '</span>';
 			}
+			if ( isset( $conflicts_keys[ $setting_id ] ) ) {
+				$setting_id_key = trim( str_replace( array( '][', '[', ']' ), '-', $setting_id ), '-' );
+				echo '<a href="#TB_inline?width=600&height=550&inlineId=snapshot-' . esc_attr( $setting_id_key ) . '" class="dashicons dashicons-warning thickbox snapshot-thickbox" title="' . esc_attr( $setting_id ) . ' ' . __( 'conflicts with following snapshot(click to expend)', 'customize-snapshots' ) . '"></a>'; ?>
```
valendesigns: Change the link to:

``` PHP
$title_text = sprintf( esc_html__( '%s has potential Snapshot conflicts (click to expand)', 'customize-snapshots' ), $setting_id );
echo '<a href="#TB_inline?width=600&height=550&inlineId=snapshot-' . esc_attr( $setting_id_key ) . '" class="dashicons dashicons-warning thickbox snapshot-thickbox" title="' . esc_attr( $title_text ) . '"></a>';
```

php/class-post-type.php
```diff
@@ -380,18 +386,23 @@ public function render_data_metabox( $post ) {
...
+				<div id="snapshot-<?php echo esc_attr( $setting_id_key ); ?>" style="display:none;">
+					<?php foreach ( $conflicts_keys[ $setting_id ] as $data ) { ?>
+								<details>
```
valendesigns: The are too many tabs

php/class-post-type.php
PatelUtkarsh: By adding that we can directly append OR in `$query .= $wpdb->prepare( 'OR post_content LIKE %s', '%' . $wpdb->esc_like( wp_json_encode( $setting_id ) ) . '%' );`
However i have updated code using `implode` and removed `1 = 2`

westonruter:
Will review once #68 is merged.

php/class-customize-snapshot-manager.php
```diff
@@ -1469,6 +1470,26 @@ public function render_templates() {
 				<p>{{ data.message }}</p>
 			</div>
 		</script>
+
+		<script type="text/html" id="tmpl-snapshot-conflict">
+			<?php $title_text = sprintf( esc_html__( 'Potential Snapshot conflicts (click to expand)', 'customize-snapshots' ) );?>
```
valendesigns: `sprintf` isn't needed here the way this string is generated.

php/class-customize-snapshot-manager.php
```diff
@@ -1469,6 +1470,26 @@ public function render_templates() {
...
+			<# id= data.setting_id.replace( /]/g, '' ).split( '[' ).filter( Boolean ).join( '-' ); #>
+			<a href="#TB_inline?width=600&height=550&inlineId=snapshot-conflicts-{{id}}" class="dashicons dashicons-warning thickbox snapshot-conflicts-thickbox" title="<?php echo $title_text; ?>"></a>
+			<div id="snapshot-conflicts-{{id}}" style="display:none;">
```
valendesigns: Why do you need to add `style="display:none;"`?

PatelUtkarsh: `thickbox` is initially hidden on click of conflict icon it should open Thickbox. I will move this css to css file.

westonruter:
As for embedding the conflict button when a control is embedded, you could take this approach:

``` js
component.addConflictButton = function addConflictButton( control ) {
    control.deferred.embedded.done( function() {
        var onFirstChange, hasDirty;

        if ( ! control.setting ) {
            return;
        }

        onFirstChange = function() {
            _.each( control.settings, function( setting ) {
                setting.unbind( onFirstChange );
            } );

            // @todo Inject notification button template (even hidden) into control.container somewhere.
            // @todo Do Ajax request to obtain potential conflicts for control.settings
            // @todo The call to do the Ajax request for settings should be batched, so that when loading a dirty snapshot, all of the settings are requested in one Ajax request. See wp.customize.selectiveRefresh.requestPartial for an example of this.
        };

        hasDirty = _.find( control.settings, function( setting ) {
            return setting._dirty;
        } );
        if ( hasDirty ) {
            onFirstChange();
        } else {
            _.each( control.settings, function( setting ) {
                setting.bind( onFirstChange );
            } );
        }
    } )
};

wp.customize.control.bind( 'add', component.addConflictButton );
wp.customize.control.each( component.addConflictButton );
```

westonruter:
@lgedeon @PatelUtkarsh just to clarify, even for myself. What Luke is working on in https://github.com/xwp/wp-customize-concurrency/pull/4 is closely related but distinct from what is being worked on here.

This PR here (#71) is about there is one for showing ​_potential_​ conflicts with ​_other_​ snapshots.
Luke's https://github.com/xwp/wp-customize-concurrency/pull/4 is about _actual_ conflicts _published_ changes. Once that in place, then he'll be able to move on to showing _actual_ conflicts with _saved_ changes to the _same_ snapshot, when someone else may be editing the same snapshot in another session. That would be tackled in #60.

php/class-customize-snapshot-manager.php
```diff
@@ -1480,6 +1481,30 @@ public function render_templates() {
...
+
+		<script type="text/html" id="tmpl-snapshot-conflict">
+			<# id= data.setting_id.replace( /]/g, '' ).split( '[' ).filter( Boolean ).join( '-' ); #>
+			<div id="snapshot-conflicts-{{id}}" class="snapshot-conflict-thickbox-content">
+				    <# _.each( data.conflicts, function( setting ) { #>
+						<details>
+							<summary>
+								<code>{{setting.name}}</code>
+								<a href="{{setting.editLink}}">
+									(<?php _e( 'edit', 'customize-snapshots' ); ?>)
```
westonruter: Instead of “edit” how about “inspect”? Also, the parents should be _inside_ the translated string, and best to use `esc_html_e()` here.

js/customize-snapshots.js
```diff
@@ -791,6 +793,141 @@
...
+
+			onFirstChange = component.onConflictFirstChange = function() {
+				_.each( control.settings, function( setting ) {
+					setting.unbind( onFirstChange );
+					if ( ! control.container.find( '.snapshot-conflicts-button' ).length ) {
+						buttonTemplate = $( $.trim( component.conflict.icon( {
+							setting_id: setting.id
+						} ) ) );
+						buttonTemplate.hide();
+						control.container.find( '.customize-control-title' ).append( buttonTemplate );
```
westonruter: There is a danger here for adding a button to the title because a control may actually not have a title. However, all controls will have a… notification area. I just realized an alternative to adding a custom button like this it could be better to make use of the notification area. A notification of type `warning` could be added as depicted here: https://github.com/xwp/wp-customize-posts/issues/217#issuecomment-239746179

php/class-post-type.php
```diff
@@ -675,4 +711,68 @@ public function hide_disabled_publishing_actions( $post ) {
...
+			if ( empty( $content ) || ! is_array( $content ) ) {
+				return $return;
+			}
+			$settings = array_keys( $content );
+			if ( empty( $settings ) ) {
+				return $return;
+			}
+		}
+		$return = array();
+		$query = $wpdb->prepare( "SELECT ID, post_title, post_status, post_content FROM $wpdb->posts WHERE post_type = %s AND post_status IN ( 'draft', 'pending', 'future' ) ", static::SLUG );
```
PatelUtkarsh: @westonruter @valendesigns which post_status should we check for conflict? Currently, I am checking all 3 `'draft', 'pending', 'future'` I think we should remove `pending` but should we check `draft` or not?

westonruter: @PatelUtkarsh I think only `future` or `pending`. If you check `draft` then the conflict should be “informative” rather than a warning.

js/customize-snapshots.js
```diff
@@ -835,6 +837,149 @@
...
+					control.remove();
+				} );
+				if ( component.conflict._currentRequest ) {
+					component.conflict._currentRequest.abort();
+					component.conflict._currentRequest = null;
+				}
+				component.conflict.controls = {};
+				component.conflict.pendingRequest = {};
+
+				wp.customize.control.each( function( control ) {
```
westonruter: Use `api` instead of `wp.customize` in this case and the other cases in this PR.

php/class-post-type.php
```diff
@@ -380,18 +386,28 @@ public function render_data_metabox( $post ) {
...
-
+			if ( isset( $conflicts_settings[ $setting_id ] ) ) {
+				$setting_id_key = trim( str_replace( array( '][', '[', ']' ), '-', $setting_id ), '-' );
+				$title_text = sprintf( esc_html__( '%s has potential Snapshot conflicts (click to expand)', 'customize-snapshots' ), $setting_id );
+				echo '<a href="#TB_inline?width=600&height=550&inlineId=snapshot-' . esc_attr( $setting_id_key ) . '" class="dashicons dashicons-warning thickbox snapshot-thickbox" title="' . $title_text . '"></a>'; ?>
+				<div id="snapshot-<?php echo esc_attr( $setting_id_key ); ?>" style="display:none;">
+					<?php foreach ( $conflicts_settings[ $setting_id ] as $data ) { ?>
+						<details>
+							<summary>
+								<code><?php echo $data['name'] ?></code>
```
westonruter: Missing `esc_html()` here.

php/class-post-type.php
```diff
@@ -380,18 +386,28 @@ public function render_data_metabox( $post ) {
...
+								<a href="<?php echo esc_url( $data['editLink'] ); ?>">
+									(<?php _e( 'edit', 'customize-snapshots' ); ?>)
```
westonruter: Best to use `esc_html_e()` here instead of `_e()`. Also, I think the parens should be included in the translated string.

php/class-post-type.php
```diff
@@ -380,18 +386,28 @@ public function render_data_metabox( $post ) {
...
+								</a>
+							</summary>
+							<?php echo $this->get_printable_setting_value( $data['value'] ); ?>
```
westonruter: Should the `customize_snapshot_value_preview` filter apply on this as well?

php/class-post-type.php
```diff
@@ -675,4 +711,68 @@ public function hide_disabled_publishing_actions( $post ) {
...
+				}
+				foreach ( $conflicts_keys as $conflicts_key ) {
+					if ( ! isset( $return[ $conflicts_key ] ) ) {
+						$return[ $conflicts_key ] = array();
+					}
+					$return[ $conflicts_key ][] = array(
+						'ID' => $item['ID'],
+						'value' => $data[ $conflicts_key ]['value'],
+						'name' => $item['post_title'],
+						'editLink' => get_edit_post_link( $item['ID'], 'raw' ),
```
westonruter: PHP keys should be `snake_case` I think. So `id` instead of `ID` and `edit_link` instead of `editLink`.

php/class-post-type.php
```diff
@@ -675,4 +711,68 @@ public function hide_disabled_publishing_actions( $post ) {
...
+			$return = array();
+			if ( empty( $content ) || ! is_array( $content ) ) {
+				return $return;
+			}
+			$settings = array_keys( $content );
+			if ( empty( $settings ) ) {
+				return $return;
+			}
+		}
+		$return = array();
```
westonruter: I think `$return` should be renamed to something more descriptive, like `$conflicted_settings`.

php/class-post-type.php
```diff
@@ -675,4 +711,68 @@ public function hide_disabled_publishing_actions( $post ) {
...
+
+	/**
+	 * Get conflicts settings
+	 *
+	 * @param \WP_Post $post post to compare conflict values.
+	 * @param array    $settings setting to search for optional.
+	 *
+	 * @return array
+	 */
+	public function get_conflicts_setting( $post, $settings = array() ) {
```
westonruter: Maybe `get_conflicted_settings()` would be a better method name?

php/class-post-type.php
```diff
@@ -675,4 +711,68 @@ public function hide_disabled_publishing_actions( $post ) {
...
+		global $wpdb;
+		if ( $post ) {
```
westonruter: Either the post type for `$post` should be checked to be `customize_snapshot` or the a `Customize_Snapshot` could be passed as the argument instead.

php/class-post-type.php
```diff
@@ -316,7 +316,13 @@ public function preserve_post_name_in_insert_data( $post_data, $original_post_da
 	 */
 	public function render_data_metabox( $post ) {
 		$snapshot_content = $this->get_post_content( $post );
-
+		if ( 'publish' !== get_post_status( $post ) ) {
+			$conflicts_settings = $this->get_conflicts_setting( $post );
+			add_thickbox();
+			wp_enqueue_style( 'snapshot-admin' );
```
westonruter: Is this not too late? If it is after `wp_head` does this not mean enqueued styles miss their chance to get printed?

php/class-customize-snapshot-manager.php
```diff
@@ -1579,4 +1616,51 @@ public function override_post_date_default_data( \WP_Post &$post ) {
...
+		} elseif ( empty( $this->current_snapshot_uuid ) ) {
+			status_header( 400 );
+			wp_send_json_error( 'invalid_customize_snapshot_uuid' );
+		}
+
+		if ( isset( $_POST['setting_ids'] ) ) { // WPCS: input var ok.
+			$setting_ids = array_map( function( $key ) {
+				// Credit: http://stackoverflow.com/a/1176923/1138341.
+				return preg_replace( '/[\x00-\x1F\x80-\xFF]/', '', $key );
+			}, $_POST['setting_ids'] );
```
westonruter: Should do `wp_unslash()` on `$_POST['setting_ids']`.

php/class-customize-snapshot-manager.php
```diff
@@ -1506,6 +1508,41 @@ public function render_templates() {
...
+		<script type="text/html" id="tmpl-snapshot-notification-template">
+			<ul>
+				<# _.each( data.notifications, function( notification ) { #>
+					<li class="notice notice-{{ notification.type || 'info' }} {{ data.altNotice ? 'notice-alt' : '' }}" data-code="{{ notification.code }}" data-type="{{ notification.type }}">{{{ notification.message || notification.code }}}</li>
+				<# } ); #>
+			</ul>
+		</script>
+
+		<script type="text/html" id="tmpl-snapshot-conflict">
+			<# id= data.setting_id.replace( /]/g, '' ).split( '[' ).filter( Boolean ).join( '-' ); #>
```
westonruter: Why not just use `data.setting_id` as-is?

php/class-customize-snapshot-manager.php
```diff
@@ -1506,6 +1508,41 @@ public function render_templates() {
 				<p>{{ data.message }}</p>
 			</div>
 		</script>
+
+		<script type="text/html" id="tmpl-snapshot-conflict-button">
+			<?php $title_text = __( 'click to expand', 'customize-snapshots' ); ?>
+			<# id= data.setting_id.replace( /]/g, '' ).split( '[' ).filter( Boolean ).join( '-' ); #>
+			<span>
+				<?php esc_html_e( 'Snapshot conflicts', 'customize-snapshots' ); ?>
+				<a href="#TB_inline?width=600&height=550&inlineId=snapshot-conflicts-{{id}}" class="dashicons dashicons-warning thickbox snapshot-conflicts-button" title="<?php echo esc_attr( $title_text ); ?>"></a>
```
westonruter: This string: `#TB_inline?width=600&height=550&inlineId=snapshot-conflicts-{{id}}` should be entity-escaped (e.g. `&amp;`)

js/customize-snapshots.js
```diff
@@ -835,6 +837,149 @@
...
+								setting_id: key,
+								conflicts: value
+							} ) ) );
+							multiple = false;
+							controls = component.conflict.pendingRequest[key];
+							buttonTemplate = $( $.trim( component.conflict.icon( {
+								setting_id: key
+							} ) ) );
+							_.each( controls, function( control ) {
+								control.notificationsTemplate = component.conflict.notificationTemplate;
```
westonruter: Is `control.notificationsTemplate = component.conflict.notificationTemplate` doing anything yet?

PatelUtkarsh: Yes we are overriding `control.notificationsTemplate` so it allows markup as a notification message.

php/class-post-type.php
PatelUtkarsh: Yes moving filter inside `get_printable_setting_value`
On a second thought, we won't have filter param values inside that.

php/class-customize-snapshot-manager.php
PatelUtkarsh: It will go to html tag's `id` attribute and it can not have `[]` which setting id will have.

mohdsayed:
tested, working as expected 👍

php/class-customize-snapshot-manager.php
westonruter: Why can't an HTML `id` attribute have brackets? There are no restrictions on what an `id` can have, except it cannot have spaces. You may be recalling HTML4 or XML which restricted the characters which can be in an ID. But browsers were always forgiving here, and HTML5 says the `id` attribute can contain anything. See https://www.w3.org/TR/html5/dom.html#the-id-attribute

image

westonruter:
@PatelUtkarsh there are conflicts now in this PR. Would you please resolve?

php/class-customize-snapshot-manager.php
PatelUtkarsh: Thank you for pointing that out I was in impression that it can not have `[]` i will update the code and fix this.

valendesigns:
@PatelUtkarsh Is this PR done, or are you waiting for feedback?

PatelUtkarsh:
@valendesigns This PR is ready, Waiting for review.

js/customize-snapshots.js
```diff
@@ -889,6 +891,176 @@
...
+	component.handleConflictRequest = function conflictRequest( setting ) {
+		if ( component.conflict._currentRequest ) {
+			component.conflict._currentRequest.abort();
+			component.conflict._currentRequest = null;
+		}
+		if ( component.conflict._debouncedTimeoutId ) {
+			clearTimeout( component.conflict._debouncedTimeoutId );
+			component.conflict._debouncedTimeoutId = null;
+		}
+		component.conflict.pendingRequest[setting.id] = setting.findControls();
```
PatelUtkarsh: @westonruter Sometimes `setting.findControls();` throws an error. 

`customize-post-section.js?ver=0.7.1-alpha:327Uncaught TypeError: Cannot read property 'call' of undefined` 
Last call trace https://github.com/xwp/wordpress-develop/blob/4.5/src/wp-includes/js/customize-base.js#L565-L567

PatelUtkarsh (commented):
Needs fix before merge putting PR in WIP again

js/customize-snapshots.js
PatelUtkarsh: Okay, that was because I was running this on WP 4.5.x. 
We need to safely fall when the version is below 4.6.

valendesigns:
@PatelUtkarsh Please resolve conflicts again when you get a chance. I'll test and review after. Also update the `dev-lib` which fixes the coverage reports not working since March.

westonruter:
I haven't had the chance to review this yet, but I did want to bring up some related work that @aaronjorbin brought to my attention. In particular, the whole “[post forking](https://github.com/post-forking/post-forking)” project is very similar conceptually to changeset forking and merging. 

While the post forking project has largely ceased, he noted the merging library worked really well:  https://github.com/post-forking/post-forking/blob/master/includes/diff3.php

What's more is that I notice this makes use of CodeMirror, and CodeMirror is now in core. That means that Snapshots could provide a changeset merging interface now like this: https://codemirror.net/demo/merge.html

PatelUtkarsh:
@westonruter One issue this would have is, Snapshot allows merging multiple posts so for example if we are merging 10 posts and if 1 particular setting have conflicts with 4 post then showing in this way would be difficult. 
Let me know your thoughts.

westonruter:
@PatelUtkarsh good point! Using CodeMirror in merging would only be suitable when merging 2 changesets, then. It's probably not necessary since you've developed a separate UI for it which is probably better and more tailored for this application anyway.

PatelUtkarsh:
Todo:

- [x] Saved changeset setting should show conflicts when loaded. Check for `setting.findControls()` code.
- [x] Thickbox z-index issue. 
- [x] Current changeset should be skipped from the comparison from MySQL query as it's being added via JS.
- [x] Above Should Fix unit-test else check unit-test
- [x] Admin CSS updates and test.
- [x] Test.

cc: @sayedtaqui

PatelUtkarsh:
I think this is ready for review. @sayedtaqui

westonruter:
I plan to finally review this after WCUS!

php/class-customize-snapshot-manager-compat.php
```diff
@@ -49,6 +49,7 @@ public function enqueue_controls_scripts() {
 			'initialServerDate' => current_time( 'mysql', false ),
 			'initialServerTimestamp' => floor( microtime( true ) * 1000 ),
 			'previewingTheme' => isset( $preview_url_query_vars['theme'] ) ? $preview_url_query_vars['theme'] : '',
+			'conflictNonce' => wp_create_nonce( Post_Type::SLUG . '_conflict' ),
```
westonruter: Should this not get exported via the `customize_refresh_nonces` filter?

php/class-customize-snapshot-manager.php
```diff
@@ -640,6 +645,62 @@ public function render_templates() {
 
 			<# } #>
 		</script>
+
+		<script type="text/html" id="tmpl-snapshot-conflict-button">
+			<?php $title_text = __( 'has potential conflicts (click to expand)', 'customize-snapshots' ); ?>
```
westonruter: This will be problematic to translate.

php/class-customize-snapshot-manager.php
```diff
@@ -640,6 +645,62 @@ public function render_templates() {
...
+								{{{setting.value}}}
+							</article>
+						</details>
+					<# }); #>
+			</div>
+		</script>
+
+		<script type="text/html" id="tmpl-snapshot-conflict-value">
+			<# if ( _.isEmpty( data.value ) ) { #>
+					<em><?php esc_html_e( '(Empty String)', 'customize-snapshots' ); ?></em>
```
westonruter: Indentation is off here a bit.

php/class-customize-snapshot-manager.php
```diff
@@ -640,6 +645,62 @@ public function render_templates() {
...
+			<# } else if ( _.isString( data.value ) ||  _.isNumber( data.value ) ) { #>
+					<p>{{data.value}}</p>
+			<# } else if ( _.isBoolean(data.value) ) {
+				var temp = JSON.stringify(data.value); #>
+					<p>{{temp}}</p>
+			<# } else {
+				var temp = JSON.stringify( data.value, null, 4 ); #>
```
westonruter: Why the `temp` variable here and above? You should be able to do just `{{ JSON.stringify( data.value, null, 4 ) }}`

js/customize-snapshots.js
```diff
@@ -265,6 +285,214 @@
...
+				}
+				snapshot.conflict.controlsWithPendingRequest = {};
+			} );
+		},
+
+		/**
+		 * Update markup with current value
+		 *
+		 * @param {string} settingId setting id.
+		 * @param {object} selector control container or conflict markup container.
```
westonruter: jsdoc should have the description starting with capital. A hyphen can separate the param name from description:

```
@param {jQuery} selector - Control container or conflict markup container.
```

php/class-post-type.php
```diff
@@ -447,41 +451,82 @@ public function render_data_metabox( $post ) {
...
-			} elseif ( is_string( $value ) || is_numeric( $value ) ) {
-				$preview = '<p>' . esc_html( $value ) . '</p>';
-			} elseif ( is_bool( $value ) ) {
-				$preview = '<p>' . wp_json_encode( $value ) . '</p>';
-			} else {
-				$preview = sprintf( '<pre class="pre">%s</pre>', esc_html( Customize_Snapshot_Manager::encode_json( $value ) ) );
+			if ( isset( $conflicts_settings[ $setting_id ] ) ) {
+				$setting_id_key = str_replace( '[', '\\[', $setting_id );
+				$setting_id_key = str_replace( ']', '\\]', $setting_id_key );
+				$title_text = sprintf( '%s ' . __( 'has potential conflicts (click to expand)', 'customize-snapshots' ), $setting_id );
```
westonruter: This is problematic to translate

php/class-post-type.php
```diff
@@ -447,41 +451,82 @@ public function render_data_metabox( $post ) {
...
 		echo '</ul>';
 	}
 
+	/**
+	 * Get printable setting value
+	 *
+	 * @param mixed         $value Value to be printed.
+	 * @param string        $setting_id setting id.
+	 * @param array         $setting_params param raw array.
+	 * @param /WP_Post|null $post Post object of where setting belongs.
```
westonruter: A backslash should be used instead of a forward slash. Also, the param descriptions should be aligned vertically.

php/class-post-type.php
```diff
@@ -1082,4 +1127,69 @@ public function show_publish_error_admin_notice() {
 		</div>
 		<?php
 	}
+
+	/**
+	 * Get conflicts settings
+	 *
+	 * @param \WP_Post $post post to compare conflict values.
+	 * @param array    $settings setting to search for optional.
```
westonruter: Align descriptions and use capital to start description. For example:

```
 * @param \WP_Post $post     Post to compare conflict values.
 * @param array    $settings Setting to search for optional.
```

php/class-post-type.php
```diff
@@ -1082,4 +1127,69 @@ public function show_publish_error_admin_notice() {
...
+			if ( empty( $content ) || ! is_array( $content ) ) {
+				return $conflicted_settings;
+			}
+			$settings = array_keys( $content );
+			if ( empty( $settings ) ) {
+				return $conflicted_settings;
+			}
+		}
+		$query = $wpdb->prepare( "SELECT ID, post_name, post_title, post_status, post_content FROM $wpdb->posts WHERE post_type = %s AND post_status IN ( 'pending', 'future' ) ", static::SLUG );
+		// Todo: finalize post_status to check in.
```
westonruter: Is this todo done?

PatelUtkarsh: Do we want to keep draft or pending and future are fine? This will filter post which we will be searching conflicted settings in. @westonruter

php/class-post-type.php
mohdsayed: @westonruter Did you mean, use `_x()` because the same string has been used before? https://github.com/xwp/wp-customize-snapshots/pull/71#discussion_r155712820
or using `(` `)` is problematic?

westonruter: What if a language is Object-Verb-Subject order? The placeholder should be inside the translation string.

php/class-post-type.php
westonruter: Limiting to `pending` and `future` make sense to me. I could also see `draft` being included.  Maybe include `draft` for now and then we can remove it if it gets too noisey?

php/class-customize-snapshot-manager.php
westonruter: What if a language is Object-Verb-Subject order? The placeholder should be inside the translation string.

php/class-post-type.php
mohdsayed: Great point @westonruter  💯

js/customize-snapshots.js
```diff
@@ -273,8 +299,216 @@
...
+				} else {
+					bindFirstChange = true;
+				}
+				_.each( control.settings, function( setting ) {
+					if ( bindFirstChange ) {
+						setting.bind( function() {
+							snapshot.handleConflictRequestOnFirstChange( control );
+						} );
+					}
+					setting.bind( updateCurrentValue );
```
westonruter: I see that `addConflictButton` is getting re-called every time a `saved` happens, and here it is re-binding the `updateCurrentValue` handler to the setting. So that means this `updateCurrentValue` is going to get added again to the setting every time the changeset is saved. Shouldn't this be prevented?

js/customize-snapshots.js
```diff
@@ -273,8 +299,216 @@
...
+				}
+
+				updateCurrentValue = function() {
+					_.each( control.settings, function( setting ) {
+						snapshot.updateConflictValueMarkup( setting.id, control.container );
+					} );
+				};
+
+				hasDirty = _.find( control.settings, function( setting ) {
+					return setting._dirty;
```
westonruter: This should probably check if `setting.extended( api.Setting )`

js/customize-snapshots.js
```diff
@@ -273,8 +299,216 @@
...
+			control.deferred.embedded.done( function() {
+				var hasDirty, updateCurrentValue, bindFirstChange = false;
+
+				if ( ! control.setting || api.section( control.section() ).extended( api.OuterSection ) ) {
+					return;
+				}
+
+				updateCurrentValue = function() {
+					_.each( control.settings, function( setting ) {
+						snapshot.updateConflictValueMarkup( setting.id, control.container );
```
westonruter: This should make sure that `setting.extended( api.Setting )`

js/customize-snapshots.js
```diff
@@ -273,8 +299,216 @@
...
+				} );
+
+				if ( hasDirty ) {
+					snapshot.handleConflictRequestOnFirstChange( control );
+				} else {
+					bindFirstChange = true;
+				}
+				_.each( control.settings, function( setting ) {
+					if ( bindFirstChange ) {
+						setting.bind( function() {
```
westonruter: This anonymous function is being bound here, and then there seems to be an attempt to unbind elsewhere with `setting.unbind( snapshot.handleConflictRequestOnFirstChange );` but it won't work because it's not the same function. I think a different approach to doing once is needed here, for example:

```diff
diff --git a/js/customize-snapshots.js b/js/customize-snapshots.js
index cf94a3b..0082b5b 100644
--- a/js/customize-snapshots.js
+++ b/js/customize-snapshots.js
@@ -347,7 +347,6 @@
 			var snapshot = this;
 
 			_.each( control.settings, function( setting ) {
-				setting.unbind( snapshot.handleConflictRequestOnFirstChange );
 				snapshot.handleConflictRequest( setting, control );
 			} );
 		},
@@ -386,10 +385,17 @@
 					bindFirstChange = true;
 				}
 				_.each( control.settings, function( setting ) {
+					var changeOnce;
 					if ( bindFirstChange ) {
-						setting.bind( function() {
+						changeOnce = function () {
 							snapshot.handleConflictRequestOnFirstChange( control );
-						} );
+							setting.unbind( changeOnce );
+						};
+						setting.bind( changeOnce );
 					}
 					setting.bind( updateCurrentValue );
 				} );
```

westonruter: But this can't get reset in `resetConflicts` since the function reference will not be available.

js/customize-snapshots.js
```diff
@@ -273,8 +299,216 @@
...
+		 * Handle conflict on first settings change.
+		 *
+		 * @param {wp.customize.Control} control - Control.
+		 * @return {void}
+		 */
+		handleConflictRequestOnFirstChange: function handleConflictRequestOnFirstChange( control ) {
+			var snapshot = this;
+
+			_.each( control.settings, function( setting ) {
+				setting.unbind( snapshot.handleConflictRequestOnFirstChange );
```
westonruter: Per below, I don't think this is going to work. There are no calls to `setting.bind( snapshot.handleConflictRequestOnFirstChange )` to unbind.

js/customize-snapshots.js
```diff
@@ -273,8 +299,217 @@
...
+				snapshot.conflict.currentRequest.abort();
+				snapshot.conflict.currentRequest = null;
+			}
+
+			snapshot.conflict.controlThickboxes = {};
+			snapshot.conflict.controlsWithPendingRequest = {};
+
+			api.control.each( function( control ) {
+				_.each( control.settings, function( setting ) {
+					setting.unbind( snapshot.handleConflictRequestOnFirstChange );
```
westonruter: Per below, I don't think this is going to work. There are no calls to `setting.bind( snapshot.handleConflictRequestOnFirstChange )` to unbind.

westonruter (changes requested):
I think these are the last changes needed.

westonruter (approved):

PR #73, opened 2016-08-12
Fix the scheduling UI

westonruter:


westonruter:
@PatelUtkarsh the dates aren't showing properly in Firefox:

firefox

It looks like the fixes in https://github.com/xwp/wp-customize-posts/pull/212 may need to be applied.

@sayedwp would you like to apply them?

PatelUtkarsh:
@sayedwp I have updated the markup and added Firefox CSS but there seems another issue because of `position: absolute` of the parent container. You can take over from here.

mohdsayed:
Fixed, I have tested on chrome, firefox and safari but not on IE ( actually I do not know how to check on IE from mac  )

firefox-fix

delawski:
Unfortunately I don't have capacity to check it out on IE over the weekend, but here's what you need to use any version of IE on Mac:
modern.ie + virtual box + gulp + browserSync with proxy setup to your WP local dev environment. Hope it helps.

jacobschweitzer:
Here is what it looks like on IE Edge:
IE Edge

And on IE 11:
IE 11

Issue #19, opened 2016-04-26
Allow snapshots to be named


westonruter:
When listing snapshots in the admin (#12), the snapshots should be shown with a given name to easily identify what is contained in the snapshot (this would be beyond any settings listed below the snapshot when expanded, as mentioned in #18). 

In addition to a name, we could make use of the `post_excerpt` to capture a “commit message” for a given snapshot.

Alternatively, we could just use the post title as the commit message and not have a separate name field at all.

valendesigns:
I think we should add the ability to modify the title and excerpt. However, we filter the excerpt in the post table to display the modified settings at a glance and adding a commit log in the excerpt would override our ad hoc one, or at the very least append to it. Something to keep in mind.

westonruter:
If we just use the title field alone for the name (commit message) that would probably be simplest.

westonruter:
I think this change would necessitate changing the new “snapshot scheduling” box to be a more general “snapshot properties“ box. In addition to the scheduled date, this box would also have the title (commit message) for the snapshot. The calendar icon could then be replaced with a the edit icon, and the existing edit icon could be replaced with a “pop-out” icon.

<img width="364" alt="customize__wordpress_develops_ _just_another_customizer-driven_site_" src="https://cloud.githubusercontent.com/assets/134745/17614192/2239710e-6019-11e6-8204-d8330536f12f.png">

/cc @PatelUtkarsh

PatelUtkarsh:
@westonruter With this change should we also change the workflow of how one can schedule snapshots?

Currently, On Creation, User first has to save snapshot and then only he can schedule a snapshot in the customizer. 
Calendar icon is not visible when there is no snapshot loaded/created.

westonruter:
@PatelUtkarsh good question. Maybe that could all stay the same, but upon the initial Save the text field would be made available to let them supply a title and re-save to make it stick?

PatelUtkarsh:
![image](https://cloud.githubusercontent.com/assets/5015489/17803609/2238f42a-6615-11e6-8188-df24be127f18.png)
@westonruter @valendesigns New UI, let me know your thoughts :)

PR #74, opened 2016-08-16
Only user with customize_publish cap can publish snapshot

PatelUtkarsh:


PatelUtkarsh:
@valendesigns Should we also hide schedule icon in customizer if a user can not publish snapshot?
If do that we might need to allow it on #19 for naming and hide just the calendar UI.

westonruter:
Yes, if the user cannot publish then the scheduling UI should be hidden, I think. It should just allow them to Submit for review (as `pending`) I should think.

valendesigns:
We should definitely close any potential unauthorized means of publishing but retain the ability to see the UI and submit as pending for review.

The client's QA team needs to test features including the scheduling UI so it should not be hidden but limited to what a normal unauthorized user would experience. That means the UI needs to handle a situation where the user doesn't have the privileges, so a save would submit for review (`pending`).

cc @PatelUtkarsh @westonruter

westonruter:
@valendesigns maybe we're saying the same thing, but contributor users cannot assign a date to their posts in the WP admin, so I should think that would apply to contributors in the customizer as well. If they can only create posts with `draft` or `pending` status then they shouldn't be able to assign future dates (and a `future` status) as well:

add_new_post_ _wordpress_develop_5_ _wordpress

PR #76, opened 2016-08-17
Allow snapshots to be named

### LINKED ISSUE #19: Allow snapshots to be named

PatelUtkarsh:
Fixes #19 

This PR branch is forked from #74 `bugfix/customize-publish`

### DIFF
--- a/css/customize-snapshots.css
+++ b/css/customize-snapshots.css
@@ -1,17 +1,17 @@
 #snapshot-preview-link,
-#snapshot-schedule-button {
+#snapshot-expand-button {
 	float: right;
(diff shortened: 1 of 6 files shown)


westonruter:
So improve.

customize__wordpress_develops

PR #77, opened 2016-08-20
Pass  (snapshot object) to customize_snapshot_save filter.

lgedeon:


westonruter:
A unit test should be added for the `save` method to verify that the snapshot object is passed as the second filter param.

westonruter:
Such a test will help guard against the filter getting inadvertently removed or changed and thus breaking plugins that make use of it.

PR #82, opened 2016-08-23
Introduce REST API previewing

westonruter:
A new bar appears at the bottom of the customizer that includes the URL for REST API request for the queried object. The URL includes the snapshot's UUID when a snapshot exists. The bar is hidden when the snapshot is not saved or when the overall customizer state is not saved, since REST API previewing requires the customizer state to be written to the DB. A button is also shown for making the API request and logging the output to the console, aside from the link to the REST API endpoint URL link that opens in a new window as well.
- [ ] Allow the bar to be expanded with the REST API response overlaying the customizer preview, where each snapshot update could show the changes in a pretty-printed JSON element.
- [ ] Add unit tests.
- [ ] Allow for multiple REST API endpoint URLs to be surfaced from the preview instead of just the queried object?

rest-api-previewing

### DIFF
--- a/css/customize-snapshots.css
+++ b/css/customize-snapshots.css
@@ -201,3 +201,54 @@
 .select2-container {
 	z-index: 500100 !important;
 }
+
+/*
+ * REST API Previewing.
+ */
+#rest-api-preview {
+	position:absolute;
+	height: 0;
+	left: 0;
+	right: 0;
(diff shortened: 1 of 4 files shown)


valendesigns:
@westonruter Could you fix the merge conflicts when you get back from break?

PR #83, opened 2016-08-24
Fix behavior of update button

### LINKED ISSUE #81: Publishing changes without first updating a snapshot causes unexpected behavior
Steps to reproduce:
1. Change site title.
2. Save snapshot (to create one).
3. Change title again.
4. Publish changes (without updating snapshot)

Bug: Snapshot Update button is not disabled and is not reverted back to Save.
Bug: Clicking Update causes an error modal to appear: “The snapshot could not be saved.”

westonruter:
@PatelUtkarsh I need your help to fix some apparent regressions that were introduced with scheduling. Namely there is logic currently in `populateSetting` that seems like it should be refactored and moved to `addButtons`.

See #81 for details on the bug needing to be fixed.

Fixes #81

### DIFF
--- a/js/customize-snapshots.js
+++ b/js/customize-snapshots.js
@@ -407,10 +407,10 @@
 
 		api.state( 'saved' ).bind( function( saved ) {
 			if ( saved && ! _.isEmpty( component.schedule.container ) ) {
-				component.data.publishDate = component.getCurrentTime();
-				component.updateSchedule();
-				component.scheduleContainerDisplayed.set( false );
 				component.data.dirty = false;
+				component.data.publishDate = component.getCurrentTime();
+				component.scheduleContainerDisplayed.set( false );
+				component.updateSchedule();
 			}
 		} );
 
--- a/js/customize-snapshots.js
+++ b/js/customize-snapshots.js
@@ -765,9 +765,11 @@
 
 			// Change update button to schedule.
 			if ( component.isFutureDate() ) {
-				save.html( component.data.i18n.scheduleButton );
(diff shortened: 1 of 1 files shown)


js/customize-snapshots.js
```diff
@@ -765,9 +765,11 @@
 
 			// Change update button to schedule.
 			if ( component.isFutureDate() ) {
-				save.html( component.data.i18n.scheduleButton );
+				save.text( component.data.i18n.scheduleButton );
+			} else if ( api.state( 'snapshot-exists' ).get() ) {
+				save.text( component.data.i18n.updateButton );
 			} else {
-				save.html( component.data.i18n.updateButton );
+				save.text( component.data.i18n.saveButton );
```
westonruter, line 772: I think all of this logic related to the buttons should be moved to `addButton`.

PatelUtkarsh: We could move this logic to `addButton` but we also want to tie this logic on date input change so I kept it separate. But yes I see the advantage of keeping all button logic in `addButton`. 
Maybe we should create a event for schedule and bind this button logic in `addButton` with that event? (using `wp.customize.state.create`)

westonruter: Yes, creating more states would be good and then listening for changes to those states in a centralized location I think is the best architecture.

PatelUtkarsh:
@westonruter I've refactored code as you suggested in https://github.com/xwp/wp-customize-snapshots/pull/83#issue-172868306 as part of PR #76 - e27ed55
Let me know if it requires any changes.

Bug #81 is also fixed in 3e26906

Issue #17, opened 2016-04-26
Allow settings to be deleted from a snapshot


westonruter:
Given a UI for listing out the available snapshots in the WP Admin (#12), if opening a snapshot in its edit post screen (see also #16), there could be a listing of the settings in that snapshot along with checkboxes that when checked, a button could be clicked to remove them from the snapshot.

westonruter:
Likewise, when calling `$snapshot->set( $setting_id, null )` this should have the effect of removing the setting from the snapshot.

valendesigns:
That isn't really needed in wp-admin. I envisioned creating a backbone collection of the setting IDs and using buttons to remove and undo. The buttons would modified the collection and dynamically update a hidden input. That way when "Save Draft" is clicked the new content is updated without those IDs.

westonruter:
@valendesigns OK, so you're thinking to not do this the WP admin edit post screen, but rather inside the Customizer itself? I think this ties into that idea we had about providing a UI to list out what the current pending changes are in the Customizer. When loading the Customizer with a snapshot, we'd need to grab all of the setting values prior to calling `preview` and then export them to JS; we likewise have the original setting values in `wp.customize.settings.settings`. Then in the UI for showing the dirty settings, all that we'd need to do is have a button listed with each setting that says “revert” and the behavior would be to override the setting's value to be whatever the original/non-snapshotted value was, and then set `_dirty` to `false`. 

The trick then would be if the user updates the snapshot, we'd also need to send a list of all of the settings that have been reverted so that when the snapshot is saved, those settings can be removed from it. This was what I was referring to with `null`. If reverted settings persist in the settings that are sent to the server when updating the snapshot, but instead of being the actual (when they are dirty) value we send `null` instead, this can be a way to indicate that the setting should be dropped from the snapshot.

valendesigns:
No, I think you misunderstood me, we're talking about two different pieces of the puzzle. I'm saying that we don't need to bootstrap the Customizer in `wp-admin` and use the `CustomizeSnapshots::set()` method to remove settings by supplying a `null` value. Instead we filter the `post_content` during save to unset the IDs we want removed.

What you mentioned above still is valid if we want to revert settings in the Customizer, but that is low priority IMO.

kienstra:
**Pull Request Opened**

Hi @westonruter,
Could you please review this [pull request](https://github.com/xwp/wp-customize-snapshots/pull/84/files)? 
[...]

PR #84, opened 2016-08-24
Issue #17 : Enable deleting settings in snapshot wp-admin/post page.

kienstra:
**Request For Code Review**

Hi @westonruter,
Could you please review this pull request for Issue #17?
- [73c5b28](https://github.com/xwp/wp-customize-snapshots/commit/73c5b28201f8f951f8c3e938405486cfb6f9a40e) : [Create link](https://github.com/xwp/wp-customize-snapshots/compare/feature/allow-deleting-settings?expand=1#diff-36a80da8dbba0bdbe840617270bb1182R365) `'Remove setting'`, which toggles to `'Restore setting'` [on click](https://github.com/xwp/wp-customize-snapshots/compare/feature/allow-deleting-settings?expand=1#diff-19c3ccdc306cf3ee7efd933155ccae35R18).
  It also [sets a hidden input field](https://github.com/xwp/wp-customize-snapshots/compare/feature/allow-deleting-settings?expand=1#diff-19c3ccdc306cf3ee7efd933155ccae35R40) for each setting to be removed.
  On saving, these appear in `$_REQUEST[ 'customize_snapshot_remove_settings' ]`.
  Filter post content  [on content_save_pre](https://github.com/xwp/wp-customize-snapshots/compare/feature/allow-deleting-settings?expand=1#diff-36a80da8dbba0bdbe840617270bb1182R117) when this is saved.
  If `$_REQUEST` has settings to be removed, [filter them out](https://github.com/xwp/wp-customize-snapshots/compare/feature/allow-deleting-settings?expand=1#diff-36a80da8dbba0bdbe840617270bb1182R722).
  Also, add [simple styling](https://github.com/xwp/wp-customize-snapshots/compare/feature/allow-deleting-settings?expand=1#diff-96bf05ff71d50822b5e71fea81a765e1R1) of this 'Remove setting' link.
  It [appears red](https://github.com/xwp/wp-customize-snapshots/compare/feature/allow-deleting-settings?expand=1#diff-96bf05ff71d50822b5e71fea81a765e1R4) when it has the text "Remove setting, and blue otherwise.

### DIFF
--- a/js/customize-snapshots-admin.js
+++ b/js/customize-snapshots-admin.js
@@ -2,18 +2,16 @@
 	'use strict';
 
 	$( function() {
-		var component, linkSelector, linkText, linkActions, dataSlug, inputName;
+		var component, $linkToRemoveOrRestore, linkActions, dataSlug, inputName;
 
 		component = {};
-		linkSelector = '.snapshot-toggle-setting-removal';
-		linkText = [ 'Remove setting', 'Restore setting' ];
+		$linkToRemoveOrRestore = $( '.snapshot-toggle-setting-removal' );
 		linkActions = [ 'remove', 'restore' ];
 		dataSlug = 'cs-action';
 		inputName = 'customize_snapshot_remove_settings[]';
 
 		component.initializeLink = function() {
-			$( linkSelector ).text( linkText[ 0 ] )
-				.data( dataSlug, linkActions[ 0 ] );
+			$linkToRemoveOrRestore.data( dataSlug, linkActions[ 0 ] );
 		};
 
 		component.initializeLink();
@@ -35,9 +33,9 @@
 			$settingDisplay = component.getSettingDisplay( $link );
 			settingId = component.getSettingId( $link );
 
-			$link.text( linkText[ 1 ] )
-				.data( dataSlug, linkActions[ 1 ] )
+			$link.data( dataSlug, linkActions[ 1 ] )
 				.after( component.constructHiddenInputWithValue( settingId ) );
+			component.changeLinkText( $link );
 			$settingDisplay.removeAttr( 'open' )
 				.addClass( 'snapshot-setting-removed' );
 		};
@@ -58,13 +56,22 @@
 			.val( settingId );
 		};
 
+		component.changeLinkText = function( $link ) {
+			var oldLinkText, newLinkText;
+			oldLinkText = $link.text();
+			newLinkText = $link.data( 'text-restore' );
+
+			$link.data( 'text-restore', oldLinkText )
+				.text( newLinkText );
+		};
+
 		component.showSettingAndChangeLinkText = function( $link ) {
 			var $settingDisplay, settingId;
 			$settingDisplay = component.getSettingDisplay( $link );
 			settingId = component.getSettingId( $link );
 
-			$link.text( linkText[ 0 ] )
-				.data( dataSlug, linkActions[ 0 ] );
+			$link.data( dataSlug, linkActions[ 0 ] );
+			component.changeLinkText( $link );
 			component.removeHiddenInputWithValue( settingId );
 			$settingDisplay.removeClass( 'snapshot-setting-removed' );
 		};
@@ -73,7 +80,7 @@
 			$( 'input[name="' + inputName + '"][value="' + settingId + '"]' ).remove();
 		};
 
-		$( linkSelector ).on( 'click', function( event ) {
+		$linkToRemoveOrRestore.on( 'click', function( event ) {
 			var $clickedLink = $( this );
 
 			event.preventDefault();
--- a/js/customize-snapshots-admin.js
+++ b/js/customize-snapshots-admin.js
@@ -53,11 +53,11 @@
 			this.showSettingAndChangeLinkText = function() {
 				$clickedLink.text( linkText[ 0 ] )
 					.data( dataSlug, linkActions[ 0 ] );
-				this.removeHiddenInputWithValue( settingId );
+				this.removeHiddenInputWithValue();
 				$settingDisplay.removeClass( 'cs-removed' );
 			};
 
-			this.removeHiddenInputWithValue = function( settingId ) {
+			this.removeHiddenInputWithValue = function() {
 				$( 'input[value="' + settingId + '"]' ).remove();
 			};
 
--- a/js/customize-snapshots-admin.js
+++ b/js/customize-snapshots-admin.js
@@ -1,4 +1,5 @@
 ( function( $ ) {
+	'use strict';
 
 	$( function() {
 		var $link, linkText, linkActions, dataSlug, initializeLink;
@@ -26,18 +27,18 @@
 			settingId = $( this ).attr( 'id' );
 
 			this.isLinkSetToRemoveSetting = function() {
-				return ( linkActions[ 0 ] === clickedLinkAction );
+				return linkActions[ 0 ] === clickedLinkAction;
 			};
 
 			this.hideSettingAndChangeLinkText = function() {
 				$clickedLink.text( linkText[ 1 ] )
 					.data( dataSlug, linkActions[ 1 ] )
-					.after( this.constructHiddenInputWithValue( settingId ) );
+					.after( this.constructHiddenInputWithValue() );
 				$settingDisplay.removeAttr( 'open' )
 					.addClass( 'cs-removed' );
 			};
 
-			this.constructHiddenInputWithValue = function( settingId ) {
+			this.constructHiddenInputWithValue = function() {
 				return $( '<input>' ).attr( {
 					'name': 'customize_snapshot_remove_settings[]',
 					'type': 'hidden'
@@ -46,7 +47,7 @@
 			};
 
 			this.isLinkSetToRestoreSetting = function() {
-				return ( linkActions[ 1 ] === clickedLinkAction );
+				return linkActions[ 1 ] === clickedLinkAction;
 			};
 
 			this.showSettingAndChangeLinkText = function() {
@@ -69,4 +70,4 @@
 		} );
 
 	} );
-}( jQuery ) );
+} )( jQuery );
--- a/js/customize-snapshots-admin.js
+++ b/js/customize-snapshots-admin.js
@@ -2,71 +2,87 @@
 	'use strict';
 
 	$( function() {
-		var $link, linkText, linkActions, dataSlug, initializeLink;
+		var component, linkSelector, linkText, linkActions, dataSlug, inputName;
 
-		$link = $( '.snapshot-toggle-setting-removal' );
+		component = {};
+		linkSelector = '.snapshot-toggle-setting-removal';
 		linkText = [ 'Remove setting', 'Restore setting' ];
 		linkActions = [ 'remove', 'restore' ];
 		dataSlug = 'cs-action';
+		inputName = 'customize_snapshot_remove_settings[]';
 
-		initializeLink = function() {
-			$link.text( linkText[ 0 ] )
+		component.initializeLink = function() {
+			$( linkSelector ).text( linkText[ 0 ] )
 				.data( dataSlug, linkActions[ 0 ] );
 		};
 
-		initializeLink();
+		component.initializeLink();
 
-		$link.on( 'click', function( event ) {
-			var $clickedLink, $settingDisplay, clickedLinkAction, settingId;
+		component.isLinkSetToRemoveSetting = function( $link ) {
+			return linkActions[ 0 ] === component.getClickedLinkAction( $link );
+		};
+
+		component.isLinkSetToRestoreSetting = function( $link ) {
+			return linkActions[ 1 ] === component.getClickedLinkAction( $link );
+		};
+
+		component.getClickedLinkAction = function( $link ) {
+			return $link.data( dataSlug );
+		};
+
+		component.hideSettingAndChangeLinkText = function( $link ) {
+			var $settingDisplay, settingId;
+			$settingDisplay = component.getSettingDisplay( $link );
+			settingId = component.getSettingId( $link );
+
+			$link.text( linkText[ 1 ] )
+				.data( dataSlug, linkActions[ 1 ] )
+				.after( component.constructHiddenInputWithValue( settingId ) );
+			$settingDisplay.removeAttr( 'open' )
+				.addClass( 'snapshot-setting-removed' );
+		};
+
+		component.getSettingDisplay = function( $link ) {
+			return $link.parents( 'details' );
+		};
+
+		component.getSettingId = function( $link ) {
+			return $link.attr( 'id' );
+		};
+
+		component.constructHiddenInputWithValue = function( settingId ) {
+			return $( '<input>' ).attr( {
+				'name': inputName,
+				'type': 'hidden'
+			} )
+			.val( settingId );
+		};
+
+		component.showSettingAndChangeLinkText = function( $link ) {
+			var $settingDisplay, settingId;
+			$settingDisplay = component.getSettingDisplay( $link );
+			settingId = component.getSettingId( $link );
+
+			$link.text( linkText[ 0 ] )
+				.data( dataSlug, linkActions[ 0 ] );
+			component.removeHiddenInputWithValue( settingId );
+			$settingDisplay.removeClass( 'snapshot-setting-removed' );
+		};
+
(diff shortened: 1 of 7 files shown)


php/class-post-type.php
```diff
@@ -675,4 +678,53 @@ public function hide_disabled_publishing_actions( $post ) {
...
+			&&
+			( static::SLUG === $post->post_type )
+			&&
+			! empty( $_REQUEST[ $key_for_settings ] )
+			&&
+			isset( $_REQUEST[ static::SLUG ] )
+			&&
+			wp_verify_nonce( $_REQUEST[ static::SLUG ], static::SLUG . '_settings' )
+			&&
+			! ( defined( 'DOING_AUTOSAVE' ) && DOING_AUTOSAVE )
```
westonruter, line 863: We should be using the new `wp_doing_ajax` function if it is available, so this can be replaced with:

``` php
! ( function_exists( 'wp_doing_ajax' ) ? wp_doing_ajax() : defined( 'DOING_AUTOSAVE' ) && DOING_AUTOSAVE )
```

This will likely yield benefits for doing unit testing without having to be in a separate Ajax unit test, since the return value of `wp_doing_ajax()` can be filtered as opposed to being a fixed constant.

westonruter: Oops. 😊 I misread the constant here. Nevermind 😄

php/class-post-type.php
```diff
@@ -675,4 +678,53 @@ public function hide_disabled_publishing_actions( $post ) {
...
+		);
+
+		if ( ! $should_filter_content ) {
+			return $content;
+		}
+
+		$setting_ids_to_unset = $_REQUEST[ $key_for_settings ];
+		$data = json_decode( $post->post_content );
```
westonruter: Instead of getting `$post->post_content` should it not be using the `$content` that is already provided?

php/class-post-type.php
```diff
@@ -675,4 +678,53 @@ public function hide_disabled_publishing_actions( $post ) {
...
+		$should_filter_content = (
+			isset( $post->post_status )
+			&&
+			( 'publish' !== $post->post_status )
+			&&
+			current_user_can( $post_type_object->cap->edit_post, $post->ID )
+			&&
+			( static::SLUG === $post->post_type )
+			&&
+			! empty( $_REQUEST[ $key_for_settings ] )
```
westonruter, line 855: Should this be checked to see if it `is_array`?

php/class-customize-snapshot-manager.php
```diff
@@ -745,6 +746,21 @@ public function enqueue_frontend_scripts() {
...
 	/**
+	 * Enqueue admin scripts.
+	 *
+	 * These files control the behavior and styling of links to remove settings.
+	 * Published snapshots can't be edited, so these files are not needed on those pages.
+	 */
+	public function enqueue_admin_scripts() {
+		global $post;
+		$handle = 'customize-snapshots-admin';
+		if ( isset( $post->post_type ) && ( Post_Type::SLUG === $post->post_type ) && ( 'publish' !== $post->post_status ) ) {
```
westonruter: Should this be looking at the `$hook` arg passed in? Or should it be checking what is returned by `get_current_screen()`?

js/customize-snapshots-admin.js
```diff
@@ -0,0 +1,73 @@
+( function( $ ) {
+	'use strict';
+
+	$( function() {
+		var $link, linkText, linkActions, dataSlug, initializeLink;
+
+		$link = $( '.cs-toggle-action' );
+		linkText = [ 'Remove setting', 'Restore setting' ];
```
westonruter: These strings need to be translated, and exported from PHP into JS.

js/customize-snapshots-admin.js
```diff
@@ -0,0 +1,73 @@
+( function( $ ) {
+	'use strict';
+
+	$( function() {
+		var $link, linkText, linkActions, dataSlug, initializeLink;
+
+		$link = $( '.snapshot-toggle-setting-removal' );
+		linkText = [ 'Remove setting', 'Restore setting' ];
```
westonruter: These strings need to be translated, and exported from PHP into JS.

php/class-post-type.php
```diff
@@ -360,6 +362,7 @@ public function render_data_metabox( $post ) {
 			echo '<li>';
 			echo '<details open>';
 			echo '<summary><code>' . esc_html( $setting_id ) . '</code> ';
+			echo '<a href="#" id="' . esc_attr( $setting_id ) . '" class="snapshot-toggle-setting-removal remove"></a>';
```
westonruter: I suggest you output the actual strings here. You can do like:

``` php
echo '<a href="#" id="' . esc_attr( $setting_id ) . '" data-text-restore="<?php esc_attr_e( 'Restore setting', 'customize-snapshots' ) ?>" class="snapshot-toggle-setting-removal remove"><?php esc_html_e( 'Remove setting', 'customize-snapshots' ) ?></a>';
```

westonruter: Then you can toggle between these strings in the JS.

js/customize-snapshots-admin.js
```diff
@@ -0,0 +1,73 @@
...
+			var $clickedLink, $settingDisplay, clickedLinkAction, settingId;
+
+			event.preventDefault();
+
+			$clickedLink = $( this );
+			$settingDisplay = $( this ).parents( 'details' );
+			clickedLinkAction = $( this ).data( dataSlug );
+			settingId = $( this ).attr( 'id' );
+
+			this.isLinkSetToRemoveSetting = function() {
```
westonruter: Humm. You're adding these functions to the jQuery object itself. Maybe better to create a `component` object that you attach the functions to, and then call its methods in a click handler instead? You can see a pattern for how to create a `component` in https://github.com/xwp/wp-customize-snapshots/blob/develop/js/customize-snapshots.js and see also how it gets exported to global scope so that the component can be interacted with by other scripts, or even by you on the console.

js/customize-snapshots-admin.js
```diff
@@ -0,0 +1,73 @@
...
+
+			this.hideSettingAndChangeLinkText = function() {
+				$clickedLink.text( linkText[ 1 ] )
+					.data( dataSlug, linkActions[ 1 ] )
+					.after( this.constructHiddenInputWithValue() );
+				$settingDisplay.removeAttr( 'open' )
+					.addClass( 'snapshot-setting-removed' );
+			};
+
+			this.constructHiddenInputWithValue = function() {
```
westonruter: Better for this function to be more of a pure function that doesn't rely on its closure scope. So it can be at the top level as `component.constructHiddenInputWithValue()` and it can take a `settingId` as its argument. It is much easier to test the function like this, and there is only ever one copy, whereas here there is a copy for each time the user clicks.

js/customize-snapshots-admin.js
```diff
@@ -0,0 +1,73 @@
...
+		dataSlug = 'cs-action';
+
+		initializeLink = function() {
+			$link.text( linkText[ 0 ] )
+				.data( dataSlug, linkActions[ 0 ] );
+		};
+
+		initializeLink();
+
+		$link.on( 'click', function( event ) {
```
westonruter: The function body should be refactored, to make the logic for the functions to not rely on `this` but be passed the context as an explicit argument. So `hideSettingAndChangeLinkText` can take an argument for the setting `details` jQuery object, for example, instead of relying on the inherited `this` context.

js/customize-snapshots-admin.js
```diff
@@ -0,0 +1,73 @@
...
+
+		$link = $( '.snapshot-toggle-setting-removal' );
+		linkText = [ 'Remove setting', 'Restore setting' ];
+		linkActions = [ 'remove', 'restore' ];
+		dataSlug = 'cs-action';
+
+		initializeLink = function() {
+			$link.text( linkText[ 0 ] )
+				.data( dataSlug, linkActions[ 0 ] );
+		};
```
westonruter, line 15: This can be eliminated in favor of putting the translated text strings inline in the rendered PHP.

js/customize-snapshots-admin.js
```diff
@@ -0,0 +1,73 @@
...
+
+			this.showSettingAndChangeLinkText = function() {
+				$clickedLink.text( linkText[ 0 ] )
+					.data( dataSlug, linkActions[ 0 ] );
+				this.removeHiddenInputWithValue();
+				$settingDisplay.removeClass( 'snapshot-setting-removed' );
+			};
+
+			this.removeHiddenInputWithValue = function() {
+				$( 'input[value="' + settingId + '"]' ).remove();
```
westonruter: Also make sure that the `name` is `customize_snapshot_remove_settings[]` right?

php/class-post-type.php
```diff
@@ -114,6 +114,7 @@ public function register() {
 		add_filter( 'display_post_states', array( $this, 'display_post_states' ), 10, 2 );
 		add_action( 'admin_notices', array( $this, 'show_publish_error_admin_notice' ) );
 		add_action( 'post_submitbox_minor_actions', array( $this, 'hide_disabled_publishing_actions' ) );
+		add_filter( 'content_save_pre', array( $this, 'filter_out_settings_if_removed_in_metabox' ), 10 );
```
westonruter, line 117: Good idea to use `content_save_pre`.

php/class-post-type.php
```diff
@@ -675,4 +678,53 @@ public function hide_disabled_publishing_actions( $post ) {
...
+		);
+
+		if ( ! $should_filter_content ) {
+			return $content;
+		}
+
+		$setting_ids_to_unset = $_REQUEST[ $key_for_settings ];
+		$data = json_decode( $post->post_content );
+		foreach ( $setting_ids_to_unset as $setting_id ) {
+			unset( $data->{ $setting_id } );
```
westonruter: The second `$associative` arg should be passed to `json_decode()` above so that it will be an array so that the familiar bracket notation can be used as opposed to this somewhat-unusual object brace notation.

westonruter:
@kienstra the functionality works well! Beyond my inline feedback, something else that should be done is a unit test on `filter_out_settings_if_removed_in_metabox`.

js/customize-snapshots-admin.js
kienstra: Thanks, @westonruter. That's a good idea to refactor the functions. This [commit](https://github.com/xwp/wp-customize-snapshots/pull/84/commits/83ad0e9ee2983c6fc7e35c6c55836e9c4242f3f0) does that. Sorry for the delay.

js/customize-snapshots-admin.js
kienstra: Thanks, @westonruter. This [commit](https://github.com/xwp/wp-customize-snapshots/pull/84/commits/83ad0e9ee2983c6fc7e35c6c55836e9c4242f3f0) attaches the functions to `component`. But I still need to export `component` object.

js/customize-snapshots-admin.js
kienstra: Thanks, this [commit](https://github.com/xwp/wp-customize-snapshots/pull/84/commits/83ad0e9ee2983c6fc7e35c6c55836e9c4242f3f0) moves `constructHiddenInputWithValue` into the `component` object.

js/customize-snapshots-admin.js
kienstra: That's a good point, @westonruter. This [commit](https://github.com/xwp/wp-customize-snapshots/pull/84/commits/83ad0e9ee2983c6fc7e35c6c55836e9c4242f3f0#diff-19c3ccdc306cf3ee7efd933155ccae35R73) ensures that the `name` is `customize_snapshot_remove_settings[]`.

js/customize-snapshots-admin.js
kienstra: Thanks, [this commit](https://github.com/xwp/wp-customize-snapshots/pull/84/commits/35d5bca1f03a9fb5cd1aef66a355b91dd30eba33#diff-36a80da8dbba0bdbe840617270bb1182R363) outputs the translated strings with PHP.

php/class-post-type.php
kienstra: Thanks, @westonruter. [This line](https://github.com/xwp/wp-customize-snapshots/pull/84/commits/35d5bca1f03a9fb5cd1aef66a355b91dd30eba33#diff-36a80da8dbba0bdbe840617270bb1182R365) in a new commit outputs the strings, and [this function](https://github.com/xwp/wp-customize-snapshots/pull/84/commits/35d5bca1f03a9fb5cd1aef66a355b91dd30eba33#diff-19c3ccdc306cf3ee7efd933155ccae35R59) toggles between the strings in JavaScript.

php/class-post-type.php
kienstra: Thanks, [this commit](https://github.com/xwp/wp-customize-snapshots/pull/84/commits/92f14c04d726faa219e31350ad2ad15b5730566e) uses the second argument of `true` to return an array.

php/class-post-type.php
kienstra: That's a good point. [This commit](https://github.com/xwp/wp-customize-snapshots/pull/84/commits/f76b4a6dba05988f250582dcd87db61902ac3ec0) corrects that.

php/class-post-type.php
kienstra: That's a good idea. This [commit](https://github.com/xwp/wp-customize-snapshots/pull/84/commits/d1487c8c233a45e2ac3a7db32780dd6a45f17d1f) adds an `is_array` call to `$should_fliter_content`.

php/class-customize-snapshot-manager.php
kienstra: Hi @weston, it looks like checking the `$hook` argument in this [admin_enqueue_scripts](https://codex.wordpress.org/Plugin_API/Action_Reference/admin_enqueue_scripts) callback should be enough.
This [commit](https://github.com/xwp/wp-customize-snapshots/pull/84/commits/68fa5ae929700ab44b31f171d69d09f97074610b) adds an expression to the conditional:
`(  'post.php' === $hook )`
The [WP_Screen object](https://codex.wordpress.org/Class_Reference/WP_Screen) that `get_current_screen()` returns has some useful properties, including:

```
id = "custom_snapshot",
base = "post",
post_type = "customize_snapshot"
```

But the conditional already checks for most of these:
`if ( ( 'post.php' === $hook ) && isset( $post->post_type ) && ( Post_Type::SLUG === $post->post_type ) && ( 'publish' !== $post->post_status ) ) {`

js/customize-snapshots-admin.js
kienstra: That's a good point. [This commit](https://github.com/xwp/wp-customize-snapshots/pull/84/commits/35d5bca1f03a9fb5cd1aef66a355b91dd30eba33#diff-36a80da8dbba0bdbe840617270bb1182R365) outputs the translated strings via PHP.

kienstra:
**Thanks For Waiting**
_I Still Need To Write PHPUnit Tests_

Hi @westonruter,
Sorry for the delay in applying your feedback here. I addressed all of your comments above, but I still need to write PHPUnit tests.

westonruter:
@kienstra no apology will be accepted! It is I who probably should apologize for being slow to review PRs 😞

php/class-post-type.php
```diff
@@ -360,6 +362,7 @@ public function render_data_metabox( $post ) {
 			echo '<li>';
 			echo '<details open>';
 			echo '<summary><code>' . esc_html( $setting_id ) . '</code> ';
+			echo '<a href="#" id="' . esc_attr( $setting_id ) . '" data-text-restore="' . esc_attr( 'Restore setting', 'customize-snapshots' ) . '" class="snapshot-toggle-setting-removal remove">' . esc_html( 'Remove setting', 'customize-snapshots' ) . '</a>';
```
PatelUtkarsh: Use translation function `esc_attr__` and `esc_html__`  instade of `esc_attr( 'Restore setting', 'customize-snapshots' )` and `esc_html( 'Remove setting', 'customize-snapshots' )`

PatelUtkarsh (changes requested):

PatelUtkarsh:
In e5accee I've resolved conflict with develop.

PatelUtkarsh:
I am picking this up as discussed with @jeffpaul and @kienstra.

PatelUtkarsh (approved):

PR #87, opened 2016-08-28
Add hooks to support customize-concurrency plugin.

lgedeon:


js/customize-snapshots.js
```diff
@@ -590,7 +590,8 @@
 			api.trigger( 'customize-snapshots-update', {
 				previewUrl: url,
 				customizeUrl: customizeUrl,
-				uuid: component.data.uuid
+				uuid: component.data.uuid,
+				response: response
```
westonruter, line 594: Good call! Adds parity with the `save` event.

php/class-customize-snapshot-manager.php
```diff
@@ -1181,6 +1186,8 @@ function( $value ) {
 			wp_send_json_error( $data );
 		}
 
+		/** This filter is documented in wp-includes/class-wp-customize-manager.php */
+		$data = apply_filters( 'customize_save_response', $data, $this );
```
westonruter: Not quite, as `$this` is `CustomizeSnapshots\Customize_Snapshot_Manager` not `WP_Customize_Manager`. So you should replace `$this` with `$this->customize_manager`.

php/class-customize-snapshot-manager.php
```diff
@@ -1125,6 +1125,11 @@ public function handle_update_snapshot_request() {
 			) );
 		}
 
+		/**
+		 * Add any additional checks before saving snapshot.
```
westonruter, line 1129: Missing `@param` tags.

php/class-customize-snapshot-manager.php
```diff
@@ -1125,6 +1125,11 @@ public function handle_update_snapshot_request() {
...
+		 */
+		do_action( 'customize_snapshot_save_before', $this->snapshot, $this->customize_manager );
```
westonruter: Since this action is specific to snapshots, I think `$this->customize_manager` should be replaced with just `$this`, the `Customize_Snapshot_Manager`.

tests/php/test-class-ajax-customize-snapshot-manager.php
```diff
@@ -526,4 +526,53 @@ function test_ajax_update_snapshot_ok_for_draft_and_pending_but_not_future() {
...
+		$this->plugin->init();
+		$this->add_setting();
+
+		$that = $this; // For PHP 5.3.
+		add_action( 'customize_snapshot_save_before', function( $test_snapshot, $test_customizer ) use ( $that ) {
+			$that->actioned_snapshot = $test_snapshot;
+			$that->actioned_customizer = $test_customizer;
+		}, 10, 2 );
+		add_filter( 'customize_save_response', function( $data, $test_snapshot_manager ) use ( $that ) {
+			$that->filtered_snapshot_manager = $test_snapshot_manager;
```
westonruter: The `filtered_snapshot_manager`, `actioned_snapshot`, and `actioned_customizer` should all be declared as `public` class variables, instead of defining them as ad hoc “expando properties”.

tests/php/test-class-ajax-customize-snapshot-manager.php
```diff
@@ -526,4 +526,53 @@ function test_ajax_update_snapshot_ok_for_draft_and_pending_but_not_future() {
...
+		$this->set_input_vars( $post_vars );
+		$this->make_ajax_call( Customize_Snapshot_Manager::AJAX_ACTION );
+
+		$manager = new Customize_Snapshot_Manager( $this->plugin );
+		$manager->ensure_customize_manager();
+		$manager->init();
+
+		$this->assertEquals( $manager->snapshot(), $this->actioned_snapshot );
+		$this->assertEquals( $manager->customize_manager, $this->actioned_customizer );
+		$this->assertEquals( $manager, $this->filtered_snapshot_manager );
```
westonruter: Good.

westonruter: Oh, you should probably clean these up using something like:

``` php

    /**
     * Clean up global scope.
     */
    function clean_up_global_scope() {
        parent::clean_up_global_scope();
        $this->actioned_snapshot = null;
        $this->actioned_customizer = null;
        $this->filtered_snapshot_manager = null;
    }
```

Or you could add this to a `tearDown` method.

westonruter:
Unit tests are failing because WordPress.org's SVN is down:

```
Installing WP from https://develop.svn.wordpress.org/trunk/ to /tmp/wordpress
svn: OPTIONS of 'https://develop.svn.wordpress.org/trunk': could not connect to server (https://develop.svn.wordpress.org)
dev-lib/check-diff.sh: line 336: /tmp/wordpress/src/wp-content/db.php: No such file or directory
```

westonruter:
@lgedeon 👏

PR #89, opened 2016-08-31
Fix snapshot publish issue after settings save

PatelUtkarsh:


php/class-customize-snapshot-manager.php
```diff
@@ -794,13 +794,13 @@ function( $value ) {
...
 		);
-		$result = $this->snapshot->set( $settings_data );
+		$result = $this->snapshot->set( $settings_data, array( 'skip_validation' => true ) );
 		if ( ! empty( $result['errors'] ) ) {
 			add_filter( 'customize_save_response', function( $response ) use ( $result, $that ) {
 				$response['snapshot_errors'] = $that->prepare_errors_for_response( $result['errors'] );
 				return $response;
 			} );
-			return false;
+			return false; // Todo should remove?
```
westonruter: I think the reason for having this `return false` here is for the sake of unit testing.

lgedeon:
@PatelUtkarsh It looks like this is going to block concurrency conflict checking, but maybe I can find a different way to trigger it in this case.

Actually, @westonruter I am not sure if you want to check for conflicts between what the snapshot is publishing and what has gone live since the snapshot was started. It could be really helpful in some cases but somewhat annoying in other cases.

lgedeon:
On second thought, this use case would be better if we used the warning interface that you are using between two different snapshots instead of the block method that I use inside the same snapshot. So we can probably leave this code as is.

[Update: Removed]

westonruter:
@lgedeon 

> It looks like this is going to block concurrency conflict checking, but maybe I can find a different way to trigger it in this case.

Actually, it shouldn't. Note that it's only bypassing the validation when transitioning a snapshot to publish immediately after the customizer settings have been saved. So this is the situation where you're in the customizer, you have a snapshot there, and you hit publish. The validation is going to run as normally for the customizer save process. What we're doing here is bypassing the validation from happening a second time at `customize_save_after`. So it should still work fine.

Issue #41, opened 2016-06-09
Allow a snapshot to be copied/forked


westonruter:

PatelUtkarsh:
@westonruter When we say forked should the original post remember it's forked snapshots? (If original posts remember it's child forks then every child will need to remember it's parent as when we delete a post we need to clear that id.) 

Would there be UI that says these snapshots are forked from currently viewing snapshot? (maybe a metabox?)

Also, this will show conflicts with forked snapshot #71, Should we ignore fork conflict?

westonruter:
@PatelUtkarsh good question. I think we could leverage the `post_parent` field to capture the relationship between a forked snapshot and its original snapshot. If the parent gets deleted then the ID will either point to nothing or a trashed post, which should have the same result: the parent should be considered empty.

There could be a UI that shows snapshots that lists out the snapshots that were forked from a given snapshot (any post which has a `post_parent` matching the current post), and showing the snapshot that it was forked from (the `post_parent`, if any).

Good question on whether conflicts should be shown when they are related as a fork. I'm guessing so, but since `draft` is no longer included among the snapshots to consider I think this is fine. I think it would be unlikely to fork a `pending` or `future` snapshot. In fact, maybe forking should be disabled when the snapshot is in a non-`draft` state.

PatelUtkarsh:
@westonruter what would be the UI for showing forked snapshot meta-box?

westonruter:
@PatelUtkarsh not sure. But perhaps it could be tied into the Revisions metabox. In addition to the revisions of that specific snaphot, it could have sections that list the forked snapshots as well as the parent snapshot (if any).

PatelUtkarsh:
Revisions box will only display if the post has revisions and I couldn't find any hook that we can tie our template in.
Current UI: #90 
![image](https://cloud.githubusercontent.com/assets/5015489/18272029/1c2bcee4-7453-11e6-8bd2-48b2c1aca57e.png)

PatelUtkarsh:
@westonruter Should we go with `Fork` word only? Currently we are using `fork/copy` as button and metabox title.

@sayedwp any input on above UI?

westonruter:
“Clone” or “Duplicate” would also work. But I think “Fork” works well for now since it has a distinct history that would not be entailed by other words.

jeffpaul:
[...]

PR #90, opened 2016-09-05
Add snapshot fork feature

PatelUtkarsh:
Fixes: #41 
Any UI improvements are welcome. 

Remaining:
- [x] Post meta Meta fork/copy 
- [x] Should append (fork) in a post title or not confirm and update.
- [x] Unit-test


PatelUtkarsh:
@westonruter / @valendesigns This PR is ready for review.
This will conflict with #84 and I will be happy to resolve conflicts.

js/customize-snapshots-admin.js
```diff
@@ -0,0 +1,35 @@
+/* global jQuery */
+(function( $, wp ) {
+	'use strict';
+	var component = {};
+
+	component.forkClick = function() {
+		var $forkButton = $( '#snapshot-fork' ),
+			$forkSnipper = $( '.snapshot-fork-spinner' ),
```
valendesigns: We could change this to `$spinner` or at the very least fix the typo and change it to `$forkSpinner`.

js/customize-snapshots-admin.js
```diff
@@ -0,0 +1,35 @@
...
+			$forkList = $( '#snapshot-fork-list' ),
+			forkItemTemplate = wp.template( 'snapshot-fork-item' );
```
valendesigns: Add return space.

js/customize-snapshots-admin.js
```diff
@@ -0,0 +1,35 @@
...
+		$forkButton.on( 'click', function( e ) {
+			var request;
+			$forkSnipper.addClass( 'is-active' );
+			request = wp.ajax.post( 'snapshot_fork', {
+				ID: $forkButton.data( 'post-id' ),
+				nonce: $forkButton.data( 'nonce' )
+			} );
```
valendesigns: Add return space.

js/customize-snapshots-admin.js
```diff
@@ -0,0 +1,35 @@
...
+			request.always( function() {
+				$forkSnipper.removeClass( 'is-active' );
+			} );
```
valendesigns: Add return space.

js/customize-snapshots-admin.js
```diff
@@ -0,0 +1,35 @@
...
+			request.done( function( data ) {
+				var item;
+				item = $( $.trim( forkItemTemplate( data ) ) );
+				$forkList.append( item );
+			} );
```
valendesigns: Add return space.

php/class-customize-snapshot-manager.php
```diff
@@ -747,6 +748,23 @@ public function enqueue_frontend_scripts() {
...
+	 * Enqueue admin scripts.
+	 *
+	 * These files control the behavior and styling of links to remove settings.
+	 * Published snapshots can't be edited, so these files are not needed on those pages.
+	 *
+	 * @param String $hook Current page.
+	 */
+	public function enqueue_admin_scripts( $hook ) {
+		global $post;
+		$handle = 'customize-snapshots-admin';
```
valendesigns: This could be moved into the conditional, and add return space after the global.

php/class-post-type.php
```diff
@@ -341,11 +353,15 @@ public function render_data_metabox( $post ) {
 
 			$frontend_view_url = get_permalink( $post->ID );
 			echo sprintf(
-				'<a href="%s" class="button button-secondary">%s</a>',
+				'<a href="%s" class="button button-secondary">%s</a> ',
```
valendesigns, line 423: Is the extra whitespace in this line change a typo?

php/class-post-type.php
```diff
@@ -322,6 +332,8 @@ public function render_data_metabox( $post ) {
 		echo sprintf( '%1$s %2$s %3$s', esc_html__( 'Modified:', 'customize-snapshots' ), esc_html( get_the_modified_date( '' ) ), esc_html( get_the_modified_time( '' ) ) ) . '<br>';
 		echo '</p>';
 
+		$fork_markup = '<a href="#" id="snapshot-fork" class="button button-secondary" data-post-id="' . $post->ID . '" data-nonce="' . wp_create_nonce( 'snapshot-fork' ) . '">' . esc_html__( 'Fork', 'customize-snapshots' ) . '</a>';
```
valendesigns: This should be a button not a link and also using `sprintf` would clean things up. Something like:

```
$fork_markup = sprintf(
    '<button id="snapshot-fork" class="button button-secondary" data-post-id="%s" data-nonce="%s">%s</button><span class="spinner snapshot-fork-spinner"></span>',
    $post->ID,
    wp_create_nonce( 'snapshot-fork' ),
    esc_html__( 'Fork', 'customize-snapshots' )
);
```

php/class-post-type.php
```diff
@@ -416,6 +432,30 @@ public function render_data_metabox( $post ) {
...
+		<ul id="snapshot-fork-list"><?php
+		if ( $post_query->have_posts() ) {
+			foreach ( $post_query->get_posts() as $p ) {
+				echo '<li><a href="' . esc_url( get_edit_post_link( $p ), 'raw' ) . '">' . get_the_title( $p ) . '</a></li>';
+			}
+		} ?>
+		</ul><?php
+		if ( $post->post_parent ) {
+			$parent = get_post( $post->post_parent );
+			echo '<h2>' . __( 'Parent:', 'customize-snapshots' ) . '&nbsp;<a href="' . esc_url( get_edit_post_link( $parent, 'raw' ) ) . '">' . get_the_title( $parent ) . '</a></h2>';
```
valendesigns: Could probably remove the `&nbsp;` and just add a space.

php/class-post-type.php
PatelUtkarsh: It is a change.

valendesigns: If it's just to add visual space to the buttons then it should probably be done with CSS on `#snapshot-fork` instead.

PatelUtkarsh: Just being consistent with https://github.com/xwp/wp-customize-snapshots/blob/0c20ece186044ad94a0d381c394c15d4030b83c4/php/class-post-type.php#L349

Should I remove space between both button and add CSS?

js/customize-snapshots-admin.js
```diff
@@ -0,0 +1,39 @@
...
+
+			e.preventDefault();
+		} );
+	};
+	component.init = function() {
+		$( function() {
+			component.forkClick();
+		} );
+	};
+	component.init();
```
westonruter: This nice component isn't exposed anywhere for the world to see. I suggest `return component;` here, and to not do self-booting script. So the top of this script can be like:

``` js
var CustomizeSnapshotsAdmin = (function( $, wp ) {
```

And then when enqueueing the script, also do:

``` php
wp_add_inline_script( 'customize-snapshots-admin', 'CustomizeSnapshotsAdmin.init()' );
```

This will allow the JS to be more readily extended and tested.

php/class-post-type.php
```diff
@@ -822,4 +867,71 @@ public function admin_show_merge_error() {
...
+				<li><a href="{{data.edit_link}}">{{data.post_title}}</a></li>
+			</script>
+		<?php }
+	}
+
+	/**
+	 * Handles snapshot fork ajax request.
+	 */
+	public function handle_snapshot_fork() {
+		if ( ! check_ajax_referer( 'snapshot-fork', 'nonce', false ) ) {
```
westonruter, line 1274: This should probably do a capability check for good measure.

PatelUtkarsh:
I've updated code in f5b86b1 
Also replaced action `admin_print_footer_scripts-post.php` with `admin_footer-post.php` as ealier action was not available in WP 4.5.4.

php/class-post-type.php
```diff
@@ -822,4 +867,77 @@ public function admin_show_merge_error() {
...
+	 */
+	public function handle_snapshot_fork() {
+		$post_type_object = get_post_type_object( self::SLUG );
+		$is_authorized_request = ( ! check_ajax_referer( 'snapshot-fork', 'nonce', false )
+		                           ||
+		                           ! isset( $_POST['ID'] )
+		                           ||
+		                           ! intval( $_POST['ID'] )
+		                           ||
+		                           ! current_user_can( $post_type_object->cap->edit_post, $_POST['ID'] )
```
westonruter: `intval( $_POST['ID'] )` should be passed into the `current_user_can()` check instead of a bare `$_POST['ID']`. As it stands right now, this is passing an unsanitized post value into `current_user_can()`. For example, if `$_POST['ID']` is `'234abc'` then this will run through `intval()` as `234` and pass that check, but then it will get passed in as that unsanitized string to the `current_user_can()`.

You should also make sure that there actually exists a post for the given post and that the post type matches. Looks like you're doing that down below as well, but I think the post's existence and type should be verified before it gets run through the cap check.

You may consider breaking up this condition into multiple `if` statements and then moving `$post_id = intval( $_POST['ID'] );` up so that you can re-use `$post_id`.

westonruter (dismissed):

valendesigns:

PatelUtkarsh:
Todo:
- [x] Test
- [x] Style @sayedtaqui

PatelUtkarsh:
Looks good, Ready for review. @westonruter

westonruter:
It appears that the coveralls Composer package is broken. I can see:

> Skipped installation of bin bin/coveralls for package satooshi/php-coveralls: file not found in package

I'll disable it.

westonruter (approved):

PR #92, opened 2016-10-06
Snapshot merge

### LINKED ISSUE #67: Allow multiple snapshots to be merged into a single snapshot
Inverse of #16.

Probably should create a new snapshot post post that contains the merged settings from the other snapshots. We could use bulk actions to implement this.

PatelUtkarsh:
Fixes #67

### DIFF
--- a/php/class-post-type.php
+++ b/php/class-post-type.php
@@ -723,7 +723,7 @@ class Post_Type {
 				$compare_a = $a->post_date;
 			}
 			if ( '0000-00-00 00:00:00' === $compare_b ) {
-				$compare_a = $b->post_date;
+				$compare_b = $b->post_date;
 			}
 			return strtotime( $compare_a ) - strtotime( $compare_b );
 		} );
--- a/php/class-post-type.php
+++ b/php/class-post-type.php
@@ -115,6 +115,15 @@ class Post_Type {
 		add_action( 'admin_notices', array( $this, 'show_publish_error_admin_notice' ) );
 		add_action( 'post_submitbox_minor_actions', array( $this, 'hide_disabled_publishing_actions' ) );
 		add_action( 'admin_print_scripts-revision.php', array( $this, 'disable_revision_ui_for_published_posts' ) );
+
+		// Version check for bulk action.
+		if ( version_compare( get_bloginfo( 'version' ), '4.7', '>=' ) ) {
+			add_filter( 'bulk_actions-edit-' . self::SLUG, array( $this, 'add_snapshot_bulk_actions' ) );
+			add_filter( 'handle_bulk_actions-edit-' . self::SLUG, array( $this, 'handle_snapshot_bulk_actions' ), 10, 3 );
+		} else {
+			add_action( 'admin_print_footer_scripts-edit.php', array( $this, 'snapshot_merge_print_script' ) );
+			add_action( 'load-edit.php', array( $this, 'handle_snapshot_bulk_actions_workaround' ) );
+		}
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


php/class-post-type.php
```diff
@@ -675,4 +685,140 @@ public function hide_disabled_publishing_actions( $post ) {
...
+				$compare_a = $a->post_date;
+			}
+			if ( '0000-00-00 00:00:00' === $compare_b ) {
+				$compare_a = $b->post_date;
+			}
+			return strtotime( $compare_a ) - strtotime( $compare_b );
+		} );
+
+		foreach ( $posts as $post ) {
+			$post->post_content = $this->get_post_content( $post );
```
westonruter: I think the data parsed from the `post_content` should be put into another array instead of stuffed back into `$post`. So maybe:

``` php
$snapshots_data = array();
foreach ( $posts as $post ) {
    $snapshots_data[] = $this->get_post_content( $post );
}
```

And use `$snapshots_data` instead of `$content`.

php/class-post-type.php
```diff
@@ -675,4 +685,140 @@ public function hide_disabled_publishing_actions( $post ) {
...
+		}
+		$content = wp_list_pluck( $posts, 'post_content' );
+		$conflict_keys = call_user_func_array( 'array_intersect_key', $content );
+		$merge_value = call_user_func_array( 'array_merge', $content );
```
westonruter: Better name than `$merge_value` would be `$merged_snapshot_data`.

php/class-post-type.php
```diff
@@ -115,6 +115,16 @@ public function register() {
 		add_action( 'admin_notices', array( $this, 'show_publish_error_admin_notice' ) );
 		add_action( 'post_submitbox_minor_actions', array( $this, 'hide_disabled_publishing_actions' ) );
 		add_action( 'admin_print_scripts-revision.php', array( $this, 'disable_revision_ui_for_published_posts' ) );
+
+		// Version check for bulk action.
+		if ( version_compare( get_bloginfo( 'version' ), '4.7', '>=' ) ) {
```
westonruter, line 120: Nice. Better than using the `$wp_version` global which I usually use. I'll do this in the future.

php/class-post-type.php
```diff
@@ -675,4 +685,140 @@ public function hide_disabled_publishing_actions( $post ) {
...
+		$merge_value = call_user_func_array( 'array_merge', $content );
+
+		foreach ( $conflict_keys as $key => $conflict_val ) {
+			$original_values = array();
+			foreach ( $posts as $post ) {
+				if ( isset( $post->post_content[ $key ] ) ) {
+					$original_values[] = $post->post_content[ $key ];
+				}
+			}
+			array_pop( $original_values );
```
westonruter: Maybe leave the winner in the merge conflict intact?

php/class-post-type.php
```diff
@@ -675,4 +685,140 @@ public function hide_disabled_publishing_actions( $post ) {
...
+			$merge_value[ $key ]['merge_conflict'] = $original_values;
```
westonruter: Maybe `merge_conflict` should use the source UUID as they key for each value here? Then we'll be able to look up later which snapshot the value came from.

php/class-post-type.php
```diff
@@ -675,4 +685,140 @@ public function hide_disabled_publishing_actions( $post ) {
...
+	public function snapshot_merge_print_script() {
+		global $post_type;
+		if ( self::SLUG === $post_type ) {
+			?>
+			<script type="text/javascript">
+		      jQuery(document).ready(function() {
+			      jQuery('<option>').val('merge_snapshot').text('<?php esc_html_e( 'Merge Snapshot', 'customize-snapshots' ); ?>').appendTo("select[name='action']");
+			      jQuery('<option>').val('merge_snapshot').text('<?php esc_html_e( 'Merge Snapshot', 'customize-snapshots' ); ?>').appendTo("select[name='action2']");
+		      });
+		    </script>
```
westonruter: Indentation problem.

php/class-post-type.php
```diff
@@ -675,4 +685,140 @@ public function hide_disabled_publishing_actions( $post ) {
...
+	 * Insert script for adding merge snapshot bulk action polyfill.
+	 */
+	public function snapshot_merge_print_script() {
+		global $post_type;
+		if ( self::SLUG === $post_type ) {
+			?>
+			<script type="text/javascript">
+		      jQuery(document).ready(function() {
+			      jQuery('<option>').val('merge_snapshot').text('<?php esc_html_e( 'Merge Snapshot', 'customize-snapshots' ); ?>').appendTo("select[name='action']");
+			      jQuery('<option>').val('merge_snapshot').text('<?php esc_html_e( 'Merge Snapshot', 'customize-snapshots' ); ?>').appendTo("select[name='action2']");
```
westonruter: These lines aren't right. There could be a translation string that has an apostrophe and this would cause a JS error. I suggest refactoring as follows:

``` js
jQuery( function( $ ) {
    var optionText = <?php echo wp_json_encode( __( 'Merge Snapshot', 'customize-snapshots' ) ); ?>;
    $( 'select[name="action"], select[name="action2"]' ).each( function() {
        var option = $( '<option>', {
            text: optionText,
            value: 'merge_snapshot'
        } );
        $( this ).append( option );
    } );
} );
```

westonruter (changes requested):

php/class-post-type.php
PatelUtkarsh: I have changed that, I kept it as such because I thought we will need mapping of post and data. However, I never used that mapping, but now to add `uuid` in `merge_conflcit` i will use it, 
Creating another array like

``` php
$snapshot_post_data = array();
foreach ( $posts as $post ) {
    $snapshot_post_data[] = array(
        'data' => $this->get_post_content( $post ),
        'uuid' => $post->post_name,
    );
}
```

php/class-post-type.php
PatelUtkarsh: Super clean 👍  🙇

westonruter:
@PatelUtkarsh is this still a WIP or is it complete from your perspective?

PatelUtkarsh:
@westonruter Resolve Conflict UI part is remaining.

PatelUtkarsh:
@westonruter I am working on Resolve Conflict UI on a separate branch so this can be reviewed and merged if you think Resolve Conflict UI is not critical.

westonruter (approved):

PR #95, opened 2016-10-24
Preview future site state

### LINKED ISSUE #21: Preview future scheduled snapshots
With snapshots being scheduled (#15) there could be multiple snapshots that will have been applied by a given time. There needs to be a way to be able to indicate a future time and preview how the site will appear with all of the future-scheduled snapshots merged together, in order, and applied to the Customizer state.

PatelUtkarsh:
Fixes #21 

**Todos**
- [x] Add feature to preview site state
- [x] Add unit-test 
- [x] Add WP date-picker 
- [x] Update unit-test for date picker
- [x] Update css for date picker
- [x] Improve code coverage

Dependent on PR #93


miina:
@PatelUtkarsh Bringing the conversation over to here:

> M: When checking the future preview with 1 Snapshot (without conflicts), I could still edit it in Customizer and change a value with clicking "Schedule", is this expected behavior? Maybe it is :)
> 
> U: I haven't got approval on that one yet, So if there is only one snapshot we are directly previewing it right now in (#95), Maybe we should fork/duplicate it and preview it. I thought because it is just previewing we can skip extra steps of duplication.
> 
> M: When checking the future preview with multiple Snapshots scheduled and when I click Customize and make some edits, it's going to create a new merged Snapshot, is this expected? However, this new Snapshot will not be Scheduled unless scheduled specifically, so it won't actually make any changes. Maybe this can be confusing? Thoughts?
> 
> U: Yes, so when you preview from dashboard it creates new merge snapshot with auto-draft status(merging all future snapshots comparing to date). If you start editing it, it will put that in draft status and will allow you to edit.

Okay, thinking that it would be great if previewing a future site would have consistency. So that if I make edits in Customizer while previewing a future site, always X happens. Otherwise it's confusing in terms of UX that in case in code only a single Snapshot is found, then X happens but if more Snapshots are found, Y happens.
Also, it would be good to know somehow that I'm not previewing a Snapshot I've created but the site preview. For example when live previewing a theme, the Customizer has this button instead of the usual `Save & Publish`:
live-preview
Maybe when making edits while previewing, it'd also be useful to use a similar logic and change the button text? Thoughts?

PatelUtkarsh:
So in a case of a single snapshot we should create a duplicate snapshot to remove that inconsistency (depends on #90), I will add to-do.

About UX I am not sure, @westonruter thoughts?

westonruter:
Should previewing future state where multiple changesets/snapshots are combined have a readonly mode? Should previewing this open on the frontend with the snapshot UUID supplied rather than opening the customizer with the UUID supplied? I should think that the virtual future-merged changesets would be represented in a new `auto-draft` snapshot/changeset that would be readonly and wouldn't be used for making edits. If you did want to explicitly edit the merged state, then you could click multiple future snapshots to then merge together into a new snapshot to make that explicit. Some ideas.

PatelUtkarsh:
Yes, read-only makes sense.
In the case of the read-only preview, I've removed wp_admin_bar link for `inspect snapshot`(post-edit) link and `customize` (don't add UUID in this case).  I've done that in 1e043e9.

valendesigns:

PatelUtkarsh:
Todo:
- [x] Remove publish changeset button from admin bar when previewing future state.
- [x] Test

php/class-post-type.php
```diff
@@ -928,15 +963,24 @@ public function handle_snapshot_merge( $redirect_to, $do_action, $post_ids ) {
...
+			$compare_key_b = 'post_modified';
+			// If status is scheduled then post_date takes more priority.
+			if ( get_post_status( $a ) === 'future' ) {
+				$compare_key_a = 'post_date';
+			}
+			if ( get_post_status( $b ) === 'future' ) {
+				$compare_key_b = 'post_date';
+			}
+			$compare_a = $a->$compare_key_a;
+			$compare_b = $b->$compare_key_b;
```
PatelUtkarsh, line 985: @westonruter When you review this PR, I've made this change on snapshot merge function. 
This will give priority to scheduled post on merge. This was needed for preview future site state but this will also now affect changeset merge option.

Let know your thoughts, earlier we were merging based on `post_modified` without checking status.
Now in case of scheduled post it uses `post_date` otherwise it will use `post_modified`.

westonruter: This makes sense to me 👍

php/class-customize-snapshot-manager.php
```diff
@@ -386,11 +394,26 @@ static public function is_valid_uuid( $uuid ) {
...
-		$this->replace_customize_link( $wp_admin_bar );
-		$this->add_changesets_admin_bar_link( $wp_admin_bar );
-		$this->add_resume_snapshot_link( $wp_admin_bar );
-		$this->add_post_edit_screen_link( $wp_admin_bar );
-		$this->add_publish_changeset_link( $wp_admin_bar );
+
+		// If previewing future state of changeset only add exit changeset preview link in admin bar.
+		$wp_customize = $this->get_customize_manager();
+		$is_future_state_preview = (
+				is_customize_preview()
```
westonruter: Looks like too much indentation here.

php/class-customize-snapshot-manager.php
```diff
@@ -576,9 +599,14 @@ public function add_snapshot_exit_link( $wp_admin_bar ) {
 		if ( ! is_customize_preview() ) {
 			return;
 		}
+		$wp_customize = $this->get_customize_manager();
+		$title = __( 'Exit Changeset Preview', 'customize-snapshots' );
+		if ( $this->is_previewing_future_state_changeset( $wp_customize->changeset_post_id() ) ) {
+			$title = __( 'Exit Preview Of Future Site State', 'customize-snapshots' );
```
westonruter: “Of” should be just “of” according to English style.

php/class-dashboard-widget.php
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	/**
+	 * Render widget.
+	 */
+	public function render_widget() {
+		$date_time = current_time( 'mysql' );
+		$date_time = new \DateTime( $date_time );
+		if ( isset( $_POST['year'], $_POST['month'], $_POST['day'], $_POST['hour'], $_POST['minute'], $_REQUEST['_wpnonce'] ) && wp_verify_nonce( $_REQUEST['_wpnonce'], 'customize_site_state_future_snapshot_preview' ) ) {
+			$date_time->setTimestamp( strtotime( "{$_POST['year']}-{$_POST['month']}-{$_POST['day']} {$_POST['hour']}:{$_POST['minute']}" ) );
```
westonruter: These should probably be validated as only containing integers.

php/class-dashboard-widget.php
```diff
@@ -0,0 +1,200 @@
...
+			echo '<p class="error-message">' . esc_html( $this->error[ $this->error_code ] ) . '</p>';
+		} ?>
+		<form method="post">
+				<div class="preview-future-state date-inputs clear">
+					<label>
+						<span class="screen-reader-text"><?php esc_html_e( 'Month', 'customize-snapshots' ); ?></span>
+						<?php $month = Customize_Snapshot_Manager_Compat::get_month_choices(); ?>
+						<select id="snapshot-date-month" class="date-input month" data-date-input="month" name="month">
+							<?php
+							foreach ( $month['month_choices'] as $month_choice ) {
```
westonruter: Since we're embedding PHP in HTML, let's use the alternative control structure: 

```php
<?php foreach ( $month['month_choices'] as $month_choice ) : ?>
```

And then do:

```php
<option value="<?php echo esc_attr( ... ) ?>...
```

Instead of `echo`'ing a string.

php/class-dashboard-widget.php
```diff
@@ -0,0 +1,200 @@
...
+		$date_time = current_time( 'mysql' );
+		$date_time = new \DateTime( $date_time );
+		if ( isset( $_POST['year'], $_POST['month'], $_POST['day'], $_POST['hour'], $_POST['minute'], $_REQUEST['_wpnonce'] ) && wp_verify_nonce( $_REQUEST['_wpnonce'], 'customize_site_state_future_snapshot_preview' ) ) {
+			$date_time->setTimestamp( strtotime( "{$_POST['year']}-{$_POST['month']}-{$_POST['day']} {$_POST['hour']}:{$_POST['minute']}" ) );
+		}
+		if ( $this->error_code ) {
+			echo '<p class="error-message">' . esc_html( $this->error[ $this->error_code ] ) . '</p>';
+		} ?>
+		<form method="post">
+				<div class="preview-future-state date-inputs clear">
```
westonruter: Looks like this is indented one too many tabs.

php/class-dashboard-widget.php
```diff
@@ -0,0 +1,200 @@
...
+		if ( 'POST' !== $_SERVER['REQUEST_METHOD'] || ! isset( $_POST['customize-future-snapshot-preview'] ) ) {
+			return;
+		}
+		check_admin_referer( 'customize_site_state_future_snapshot_preview' );
+		if ( ! isset( $_POST['year'], $_POST['month'], $_POST['day'], $_POST['hour'], $_POST['minute'] ) ) {
+			$this->error_code = 1;
+			return;
+		}
+		$date = new \DateTime();
+		$date->setTimestamp( strtotime( "{$_POST['year']}-{$_POST['month']}-{$_POST['day']} {$_POST['hour']}:{$_POST['minute']}" ) );
```
westonruter: Probably should validate all these vars are digits.

php/class-post-type.php
```diff
@@ -395,6 +396,34 @@ public function render_data_metabox( $post ) {
...
+					echo get_the_title( $uuid_post );
+				} else {
+					$title = get_the_title( $uuid_post );
+					if ( empty( $title ) ) {
+						echo esc_html( $uuid_post->post_name );
+					} else {
+						echo get_the_title( $uuid_post ) . ' - ' . esc_html( $uuid_post->post_name );
+					}
+				}
+				echo ( 'future' === $uuid_post->post_status ) ? ' (Scheduled)' : '';
```
westonruter: Lacking translation for `(Scheduled)`.

php/class-post-type.php
```diff
@@ -970,22 +1014,56 @@ public function merge_snapshots( $post_ids ) {
...
+			// Store contributed snapshot id in setting.
+			$uuid = array();
+			foreach ( $snapshot_post_data as $post_data ) {
+				if ( isset( $post_data['data'][ $key ]['merged_uuid'] ) && is_array( $post_data['data'][ $key ]['merged_uuid'] ) ) {
+					$uuid = array_merge( $uuid, $post_data['data'][ $key ]['merged_uuid'] );
+				} elseif ( isset( $post_data['data'][ $key ] ) ) {
+					$uuid[] = $post_data['uuid'];
+				}
+			}
+			$merged_snapshot_data[ $key ]['merged_uuid'] = array_values( array_unique( $uuid ) );
```
westonruter: Instead of `merged_uuid` what about `merged_changeset_uuids` or `merged_changesets`?

php/class-post-type.php
```diff
@@ -970,22 +1014,56 @@ public function merge_snapshots( $post_ids ) {
...
 					);
 				}
+				if ( isset( $post_data['data'][ $key ]['merged_uuid'] ) ) {
+					$uuid = array_merge( $uuid, $post_data['data'][ $key ]['merged_uuid'] );
+				} else {
+					$uuid[] = $post_data['uuid'];
+				}
+			}
+			$values = wp_list_pluck( $original_values, 'value' );
+			$values = array_unique( $values, SORT_REGULAR );
```
westonruter: Nice, I didn't know about the second arg for `array_unique()`.

php/class-post-type.php
```diff
@@ -1331,4 +1409,134 @@ public function handle_snapshot_fork() {
...
+	 * This will add thickbox with radio button to select between conflicted setting values.
+	 *
+	 * @param string   $setting_id       setting id.
+	 * @param array    $value            setting value.
+	 * @param array    $snapshot_content snapshot post-content.
+	 * @param \WP_Post $post             Post object.
+	 */
+	public function resolve_conflict_markup( $setting_id, $value, $snapshot_content, $post ) {
+		if ( isset( $value['merge_conflict'] ) ) {
+			$setting_id_key = str_replace( ']', '\\]', str_replace( '[', '\\[', $setting_id ) );
```
westonruter: Why this? Why adding escapes?

php/class-post-type.php
```diff
@@ -1331,4 +1409,134 @@ public function handle_snapshot_fork() {
...
+			echo '</ul>';
+			echo '</div>';
+		} // End if().
+	}
+
+	/**
+	 * Change conflicted value in post_content, if they were changed in the meta box.
+	 *
+	 * @param String $content Post content to filter.
+	 * @return String $content Post content, possibly filtered.
```
westonruter: Add `@global \WP_Post $post`

php/class-post-type.php
```diff
@@ -1331,4 +1409,134 @@ public function handle_snapshot_fork() {
...
+				$snapshot_data[ $setting_key ]['selected_uuid'] = $data['uuid'];
+				foreach ( $snapshot_data[ $setting_key ]['merge_conflict'] as $conflict_data ) {
+					if ( isset( $conflict_data['uuid'], $conflict_data['value'] ) && $conflict_data['uuid'] === $data['uuid'] ) {
+						$snapshot_data[ $setting_key ]['value'] = $conflict_data['value'];
+						break;
+					}
+				}
+			}
+		}
+		$content = Customize_Snapshot_Manager::encode_json( $snapshot_data );
```
westonruter: Shouldn't this re-`wp_slash()`?

php/class-post-type.php
```diff
@@ -1331,4 +1409,134 @@ public function handle_snapshot_fork() {
...
+		return $content;
+	}
+
+	/**
+	 * Get snapshot merged uuid from post_content
+	 *
+	 * @param array $post_content snapshot data.
+	 * @return array merged snapshot.
+	 */
+	public function get_snapshot_merged_uuid( $post_content ) {
```
westonruter: Maybe this should be `get_merged_changesets()`. Let's also make it `private` as I don't see a need for it to be exposed.

php/class-post-type.php
PatelUtkarsh: Part of #93 will fix in it.

php/class-post-type.php
PatelUtkarsh: Will use `merged_changeset_uuids` part of #93 I'll fix in it.

php/class-post-type.php
PatelUtkarsh: Will add in #93.

php/class-post-type.php
PatelUtkarsh: Yep, will update in #93.

PatelUtkarsh:
```php
	/**
	 * Get snapshot merged uuid from post_content
	 *
	 * @internal
	 *
	 * @param array $post_content snapshot data.
	 * @return array merged snapshot.
	 */
	public function _get_merged_changesets( $post_content ) {
```
@westonruter How about this? 
If we make this private we need to mock that method in unit-test which is not best practice. We can cover this method under render function's unit-test.

php/class-post-type.php
PatelUtkarsh: Yes.

php/class-post-type.php
PatelUtkarsh: @westonruter 
`#TB_inline?width=600&height=550&inlineId=snapshot-post\[post\]\[1\]`
Thickbox uses jQuery, `inlineId` is selector and for `[` to work in jQuery it needs this escaping https://stackoverflow.com/a/2365005

PatelUtkarsh: https://api.jquery.com/category/selectors/
>To use any of the meta-characters ( such as  !"#$%&'()*+,./:;<=>?@[\]^`{|}~ ) as a literal part of a name, it must be escaped with with two backslashes: \\.

westonruter (approved):

PR #102, opened 2016-11-16
Integrate with changesets in WP 4.7

### LINKED ISSUE #99: Extend, integrate with and defer to changesets in WP 4.7
_This is an important issue to complete before the 4.7 release in December._

The underlying architecture in snapshots was committed to core in WordPress 4.7-alpha. As such, the redundant code in snapshots should be only run if on 4.6 or less. All of the redundant code should be extracted into one or more back-compat classes.

Also, now that the post type is `customize_changeset` as opposed to `customize_snapshot`, there should be a data migration strategy. Also, any filters that apply on `customize_snapshot_save` should also apply onto `customize_changeset_save_data`.

Whenever a `$_REQUEST['customize_snapshot_uuid']` is present we should during `plugins_loaded` make sure this gets mapped over to `customize_changeset_uuid` in `$_GET`, `$_POST`, and `$_REQUEST`.

In the end, the plugin should be reformulated to anticipate that more and more of the features of the plugin will be incorporated into core. So the more that we can facilitate turning off aspects of the code that are already in core, the better.

PatelUtkarsh:
Fixes #99 

**Snapshot Actions: **

**1 -** Hit "Publish" Button
- Publish ( Without Confirming )

**2 -** Select "Publish” from dropdown 
 - Confirm Publish.
  - If clicked again
    - Publish.

**3 -** Select “Save Draft” / “Save Pending" from status - 
- Change changesetStatus to draft / pending.
- Save as draft / pending.

**4 -** Hit “Draft” / “Pending"  button
- Save as draft / pending

**5 -** Select “Schedule” from dropdown
- Show the edit box
 - If date is of future 
   - Change changesetStatus to “future”
   - Save changeset.
   - If title/date is changed.
     - Auto Save title/date entered.
- If date is not of future
  - Show notification to change date to future.

**6 -** Auto Save happens from core
- If current status is publish
  - Save as “auto-draft” ( by default )
- If current status is pending / draft
  - Let the changeset save as the current status ( pending / draft  )
- If current status is future.
  - Save with future date
  - If the selected date is not of future
    - Save changeset to the previous future date used.
- If “Confirm Publish” is the text in button.
  - Save changeset to the current status ( pending/ draft/ future ).

**Issues:** 
- [x] When selecting “Publish” from the dropdown, the button changes to “Confirm Publish”. If you then click in into the preview or allow the autosave logic to run, then it goes ahead and publishes even though you never click the button. Additionally, the changeset UUID doesn't get removed from the URL.


PatelUtkarsh:
5 Commit with spock name are mine, Some app changed my git config name with system name.

js/compat/customize-snapshots-preview.js
```diff
@@ -60,7 +60,7 @@ var CustomizeSnapshotsPreview = (function( api, $ ) {
 		var customized = {};
 		api.each( function( setting ) {
 			if ( setting._dirty || -1 !== _.indexOf( component.data.initial_dirty_settings, setting.id ) ) {
-				customized[ setting.id ] = setting.get();
+				customized[setting.id] = setting.get();
```
westonruter: Not an issue, but technically the spaces should be retained since the bracketed item is a variable and not a literal.

js/customize-migrate.js
```diff
@@ -0,0 +1,46 @@
...
+		} );
+	};
+	component.migrate = function( nonce, limit ) {
+		var requestData, request;
+		requestData = {
+			nonce: nonce,
+			limit: limit
+		};
+		request = wp.ajax.post( 'customize_snapshot_migration', requestData );
+		request.always( function( data ) {
```
westonruter: Low concern, but what if an Ajax failure happens?

php/class-customize-snapshot-manager.php
```diff
@@ -758,24 +324,21 @@ public function enqueue_frontend_scripts() {
...
 			wp_enqueue_script( $handle );
 			wp_enqueue_style( $handle );
+			$exports = array(
+				'deleteInputName' => $this->get_post_type() . '_remove_settings[]',
+			);
+			wp_add_inline_script(
+				$handle,
+				sprintf( 'CustomizeSnapshotsAdmin.init( %s )', wp_json_encode( $exports ) ),
+				'after'
+			);
```
westonruter, line 332: ❤️

js/customize-snapshots.js
```diff
@@ -1,894 +1,971 @@
...
-				var requestData = {
-					status: 'pending'
-				};
-				event.preventDefault();
-				if ( component.snapshotTitle && component.snapshotTitle.val() ) {
-					requestData.title = component.snapshotTitle.val();
-				}
-				component.sendUpdateSnapshotRequest( requestData );
-			} );
+	api.Snapshots = api.Class.extend( {
```
westonruter: ❤️❣️❣️❣️

php/class-customize-snapshot-manager.php
```diff
@@ -33,7 +33,7 @@ class Customize_Snapshot_Manager {
 	/**
 	 * Post type.
 	 *
-	 * @var Post_Type
+	 * @var Post_Type|Post_Type_Back_Compt
```
westonruter: Typo: `s/Post_Type_Back_Compt/Post_Type_Back_Compat/`

php/class-customize-snapshot-manager.php
```diff
@@ -647,16 +267,19 @@ public function enqueue_controls_scripts() {
 
 		wp_enqueue_style( 'customize-snapshots' );
 		wp_enqueue_script( 'customize-snapshots' );
+
```
westonruter, line 270: Might as well add `$post = null;` here so that static analysis (in PhpStorm) doesn't complain about a potentially-undefined variable `$post` later.

js/compat/customize-snapshots.js
```diff
@@ -0,0 +1,465 @@
...
+/* eslint consistent-this: ["error", "snapshot"] */
+
+( function( api, $ ) {
+	'use strict';
+
+	if ( ! api.Snapshots ) {
+		return;
+	}
+
+	api.SnapshotsCompat = api.Snapshots.extend( {
```
westonruter: Beautiful.

js/compat/customize-snapshots.js
```diff
@@ -0,0 +1,465 @@
+/* global jQuery, wp */
+/* eslint consistent-this: ["error", "snapshot"] */
+
+( function( api, $ ) {
+	'use strict';
+
+	if ( ! api.Snapshots ) {
```
westonruter: I'd suggest removing this conditional. If this is not defined then it is clearly an error and better to not fail silently.

js/compat/customize-snapshots.js
```diff
@@ -0,0 +1,465 @@
...
+				snapshot.dateNotification.addClass( 'hidden' );
+			}
+		},
+
+		/**
+		 * Auto save edit box when the dates are changed.
+		 *
+		 * @return {void}
+		 */
+		autoSaveEditBox: function autoSaveEditor() {
```
westonruter: Todo?

js/compat/customize-snapshots.js
```diff
@@ -0,0 +1,465 @@
...
+		 * But unfortunately there is no such enable method.
+		 *
+		 * @return {void}
+		 */
+		resetSavedStateQuietly: function resetSavedStateQuietly() {
+			api.state( 'saved' )._value = true;
+		},
+
+		/**
+		 * Hides the future date notification used for 4.7.
```
westonruter: Since this is in a `compat` file, should this method be included here?

js/customize-snapshots.js
```diff
@@ -1,894 +1,971 @@
...
 			}
 
-			api.trigger( 'snapshots-ready', component );
-		} );
+			// Set the initial client timestamp.
+			snapshot.data.initialClientTimestamp = snapshot.dateValueOf();
 
-		api.bind( 'save', function( request ) {
+			api.bind( 'ready', function() {
+				snapshotExists = '{}' !== api.previewer.query().customized;
```
westonruter: This isn't going to tell us whether the snapshot exists. It's just going to tell us whether there are dirty changes that aren't yet written into the changeset. If you want to check it the changeset exists, you should look at `api.state( 'changesetStatus' )` to see if it is not an empty string.

js/customize-snapshots.js
```diff
@@ -1,894 +1,971 @@
...
-				urlParts;
+			api.bind( 'save', function( request ) {
 
-			// Update the UUID.
-			if ( response.new_customize_snapshot_uuid ) {
-				component.data.uuid = response.new_customize_snapshot_uuid;
-				component.previewLink.attr( 'target', component.data.uuid );
-			}
+				// Make sure that saved state is false so that Published button behaves as expected.
+				api.state( 'saved' ).set( false ); // @todo Move it to compat.
```
westonruter: What about when the changeset actually is published? This seems kinda hacky.

In order to get around hacking with the existing button, should maybe we just the original button entirely and just insert a new one entirely that we have complete control over?

westonruter: That being said, I realize this is part of the original plugin. So yeah, moving to back-compat would make sense.

js/customize-snapshots.js
```diff
@@ -1,894 +1,971 @@
...
+		 * Populate setting value from the inputs.
+		 *
+		 * @returns {void}
+		 */
+		populateSetting: function populateSetting() {
+			var snapshot = this,
+				date = snapshot.getDateFromInputs(),
+				scheduled, editControlSettings;
+
+			editControlSettings = _.extend( {}, snapshot.editControlSettings.get() );
```
westonruter, line 878: Cloning 👍

php/class-customize-snapshot-command.php
```diff
@@ -0,0 +1,57 @@
...
+	 *
+	 *      wp customize_snapshot migrate --dry-run
+	 *      wp customize_snapshot migrate
+	 *
+	 * @when after_wp_load
+	 *
+	 * @param array $arg        command args.
+	 * @param array $assoc_args assoc args.
+	 */
+	public function migrate( $arg, $assoc_args ) {
```
westonruter, line 34: I like to `unset( $arg )` for positional args that are intentionally not being used. This keeps static analysis happy.

php/class-customize-snapshot-command.php
```diff
@@ -0,0 +1,57 @@
...
+		$migrate_obj = new Migrate();
+		if ( $migrate_obj->compat ) {
+			\WP_CLI::error( 'You\'re using older WordPress version please upgrade 4.7 or above to migrate.' );
```
westonruter: Not translated, but probably not a big deal. Same below.

php/class-customize-snapshot-manager-back-compat.php
```diff
@@ -0,0 +1,1009 @@
...
+		}
+	}
+
+	/**
+	 * Determine whether the current snapshot can be previewed.
+	 *
+	 * @param Customize_Snapshot $snapshot Snapshot to check.
+	 * @return true|\WP_Error Returns true if previewable, or `WP_Error` if cannot.
+	 */
+	public function should_import_and_preview_snapshot( Customize_Snapshot $snapshot ) {
```
westonruter: This param should probably be explicitly a `Customize_Snapshot_Back_Compat` since the `saved` method is only defined on it.

php/class-customize-snapshot-manager-back-compat.php
```diff
@@ -0,0 +1,1009 @@
...
+	 * This must happen at `customize_register` priority prior to 11 which is when `WP_Customize_Nav_Menus::customize_register()` runs.
+	 * This is only relevant when accessing the Customizer app (customize.php), as this is where sections/controls matter.
+	 *
+	 * @see \WP_Customize_Nav_Menus::customize_register()
+	 */
+	public function preview_early_nav_menus_in_customizer() {
+		if ( ! is_admin() ) {
+			return;
+		}
+		$this->customize_manager->add_dynamic_settings( array_keys( $this->snapshot()->data() ) );
```
westonruter, line 429: This `Customize_Snapshot_Manager_Back_Compat` should probably override the `snapshot` method to just indicate that it returns a `Customize_Snapshot_Back_Compat`, since the `data` method is only available in that subclass. Or implement the `data` method on the base class.

php/class-customize-snapshot-manager.php
```diff
@@ -675,74 +298,17 @@ public function enqueue_controls_scripts() {
...
-			'home_url' => wp_parse_url( home_url( '/' ) ),
-			'l10n' => array(
-				'restoreSessionPrompt' => __( 'It seems you may have inadvertently navigated away from previewing a customized state. Would you like to restore the snapshot context?', 'customize-snapshots' ),
-			),
-		);
-		wp_add_inline_script(
-			$handle,
-			sprintf( 'CustomizeSnapshotsFrontend.init( %s )', wp_json_encode( $exports ) ),
+			$this->plugin->slug,
+			sprintf( 'new wp.customize.Snapshots( %s )', wp_json_encode( $exports ) ),
```
westonruter: This perfectly good and beautiful class is getting instantiated without saving a reference anywhere for us to use. It should get saved somewhere, like assigned to `wp.customize.snapshots`

php/class-customize-snapshot-manager.php
```diff
@@ -1444,6 +627,74 @@ public function render_templates() {
...
+							'option_text' => esc_attr__( 'Schedule' , 'customize-snapshots' ),
+							'alt_text' => esc_attr__( 'Scheduled' , 'customize-snapshots' ),
+						),
+						'pending' => array(
+							'option_text' => esc_attr__( 'Save Pending' , 'customize-snapshots' ),
+							'alt_text' => esc_attr__( 'Pending' , 'customize-snapshots' ),
+						),
+					),
+					'selected' => 'publish',
+					'confirm_publish_text' => esc_attr__( 'Confirm Publish', 'customize-snapshots' ),
```
westonruter: Using `esc_attr__()` isn't appropriate in all of these cases. It should just be `__()`. The escaping will be handled at the moment where the string gets actually embedded into the document. The templating engine will escape them properly.

php/class-customize-snapshot-manager.php
```diff
@@ -1652,4 +885,31 @@ public function override_post_date_default_data( \WP_Post &$post ) {
...
+	}
+
+	/**
+	 * Get customize uuid param name.
+	 *
+	 * @return string customize param name.
+	 */
+	public function get_customize_uuid_param() {
+		return constant( get_class( $this->post_type ) . '::CUSTOMIZE_UUID_PARAM_NAME' );
+	}
```
westonruter, line 916: 👍

php/class-migrate.php
```diff
@@ -0,0 +1,205 @@
...
+			'post_type' => 'customize_snapshot',
+			'no_found_rows' => false,
+			'update_post_meta_cache' => false,
+			'update_post_term_cache' => false,
+			'post_status' => array_keys( get_post_stati() ),
+			'posts_per_page' => $limit,
+			'fields' => 'ids', // We will use get_post() to fetch each posts.
+		);
+
+		if ( - 1 === $limit ) {
```
westonruter: `s/- 1/-1/`

php/class-migrate.php
```diff
@@ -0,0 +1,205 @@
...
+			);
+			if ( $setting instanceof \WP_Customize_Setting ) {
+				$post_data[ $prefixed_setting_id ]['type'] = $setting->type;
+			}
+		}
+		$maybe_updated = wp_update_post( wp_slash( array(
+			'ID' => $post->ID,
+			'post_type' => 'customize_changeset',
+			'post_content' => Customize_Snapshot_Manager::encode_json( $post_data ),
+		) ), true );
```
westonruter, line 210: Great, and since `revisions` support is added, this should result in a new entry being added.

php/class-plugin.php
```diff
@@ -54,7 +79,11 @@ public function __construct() {
 	 * @action after_setup_theme, 8
 	 */
 	public function init() {
-		$this->customize_snapshot_manager = new Customize_Snapshot_Manager( $this );
+		if ( $this->compat ) {
+			$this->customize_snapshot_manager = new Customize_Snapshot_Manager_Back_Compat( $this );
+		} else {
+			$this->customize_snapshot_manager = new Customize_Snapshot_Manager( $this );
+		}
```
westonruter, line 86: 👍

php/class-post-type.php
```diff
@@ -522,11 +491,6 @@ public function save( array $args ) {
 			),
 		);
 		if ( ! empty( $args['status'] ) ) {
-			if ( isset( $args['post_date'], $args['edit_date'], $args['post_date_gmt'] ) ) {
-				$post_arr['post_date'] = $args['post_date'];
-				$post_arr['edit_date'] = $args['edit_date'];
-				$post_arr['post_date_gmt'] = $args['post_date_gmt'];
-			}
```
westonruter, line 529: Shouldn't the `save` method here be calling `\WP_Customize_Manager::save_changeset_post()`?

westonruter (changes requested):
Your work looks very good. But I didn't really doubt that! 👍

js/customize-migrate.js
PatelUtkarsh: I will change it to `request.done` and will terminate ajax loop in case of fail.

css/customize-snapshots.css
```diff
@@ -216,4 +216,122 @@
 #customize-snapshot .snapshot-controls{
 	list-style: none;
 	margin: 0;
-}
\ No newline at end of file
+}
+
+.snapshot-future-date-notification.notice{
```
valendesigns: There should be a space between all the classes and the opening bracket. Which goes for the statements below, as well.

instance.php
```diff
@@ -51,3 +56,17 @@ function current_snapshot_uuid() {
...
+ *
+ * @return bool is compat.
+ */
+function is_back_compat() {
+	// Fix in case version contains 'src' for example 4.7-src in that case php version compare fails to compare correctly.
+	$wp_version = get_bloginfo( 'version' );
+	if ( $pos = strpos( $wp_version, 'src' ) ) {
+		$wp_version = substr( $wp_version, 0, $pos - 1 );
+	}
+	return version_compare( $wp_version, '4.7', '<' );
```
valendesigns: Why not use the global `$wp_version` instead of trying to check the string. The whole method could be rewritten like:

```PHP
function is_back_compat() {
	global $wp_version;

	return version_compare( $wp_version, '4.7', '<' );
}
```

js/customize-migrate.js
```diff
@@ -0,0 +1,53 @@
+/* global jQuery, wp */
+(function( $ ) {
+	'use strict';
+	var component = {
+		doingAjax: false,
+		postMigrationCount: 20
+	};
+	component.init = function() {
```
valendesigns: All these JS methods should have docblocks and there should be spacing between them.

php/class-migrate.php
```diff
@@ -0,0 +1,205 @@
...
+	 *
+	 * @var bool
+	 */
+	public $compat;
+
+	/**
+	 * Migrate constructor.
+	 */
+	public function __construct() {
+		$this->compat = is_back_compat();
```
valendesigns: I don't see the need for the `compat` class variable , it's not being used anywhere.

php/class-plugin.php
```diff
@@ -29,21 +29,46 @@ class Plugin extends Plugin_Base {
...
 		parent::__construct();
 	}
 
 	/**
+	 * Init migration.
+	 *
+	 * @action init
+	 */
+	public function init_migration() {
+		$this->migrate = new Migrate();
```
valendesigns: You could pass `$this` to `Migrate` and then have access to `$compat` instead of recreating it.

valendesigns (changes requested):

instance.php
westonruter: Actually this should also account for strings that include beta or RC. Basically I think replace everything after `-` inclusive.

I think it's fine to use the function instead of the global var since globals can be mutated.

js/compat/customize-snapshots.js
mohdsayed: No this is to override the `autoSaveEditBox ` method used in `api.Snapshots` because we are not doing autosave in `< 4.7`

js/customize-snapshots.js
mohdsayed: Moved to compat

php/class-customize-snapshot-manager.php
mohdsayed: Absolutely.

js/compat/customize-snapshots.js
mohdsayed: I am using `snapshot.statusButton.value.get()`  in `api.Snapshots` which is not available in `< 4.7` so wanted to override the method `toggleDateNotification`. But updated it to 

```
		/**
		 * Toggles date notification.
		 *
		 * @return {void}.
		 */
		toggleDateNotification: function showDateNotification() {
			var snapshot = this;
			if ( ! _.isEmpty( snapshot.dateNotification ) ) {
				snapshot.dateNotification.toggle( ! snapshot.isFutureDate() );
			}
		}
``` 
now.

js/compat/customize-snapshots.js
westonruter: Cool. Please add that add a comment :smile:

westonruter:
Let's move the Changesets link in the admin menu to appear alongside Customize in the Appearance dropdown: https://gist.github.com/westonruter/b2c9edb9a2ee83236dd9b2b4f177ae76#file-customize-changesets-admin-ui-php-L73-L84

In the future I want to see Customize as a top-level admin link with Changesets appearing as a submenu item below it.

westonruter:
Latest upstream improvements to WordPress-Coding-Standards have resulted in some new sniff violations in the latest Travis build.

php/class-migrate.php
```diff
@@ -0,0 +1,216 @@
...
+		wp_enqueue_script( 'customize-snapshot-migrate' );
+	}
+
+	/**
+	 * Show admin notice to migrate.
+	 */
+	public function show_migration_notice() {
+		?>
+		<div class="notice notice-error customize-snapshot-migration">
+			<p><?php esc_html_e( 'Existing snapshots need to be migrated in changesets in WordPress 4.7.', 'customize-snapshots' );
```
valendesigns: I think this should read a bit more like this:
> Existing Snapshots need to be migrated to Changesets, which was added to Core in WordPress 4.7. Click here to start the migration.

westonruter (approved):

PR #110, opened 2017-02-07
Add Customize menu to allow access user with 'customize' cap

PatelUtkarsh:
User with `customize` cap can not see customizer menu as it checks for `edit_theme_options` permission. This adds `Customize` as top menu and `Changeset` and submenu for such users.


westonruter:
Love the new top-level menu item. I would not be opposed to doing this for all users!

PatelUtkarsh:
I will create another PR to add customize menu on top for all user.

PR #111, opened 2017-02-07
Ensure that revisions get created when changesets are updated via explicit saves

westonruter:
Found a bug with 0.6.0-rc1 whereby revisions were no longer being made as expected when explicitly clicking the Save Draft button in the UI. The opt-in for creating a revision in core is currently indicated by whether or not a `customize_changeset_status` param is present. This is not ideal, but since the param is getting stripped by Customize Snapshots if the status is the same as the current status, I added a workaround for this to explicitly indicate the need to create a revision via another channel. There is probably a better solution for this, not at least to allow a `save_revision` to be passed when calling `wp.customize.previewer.save()`. 

See: https://github.com/WordPress/wordpress-develop/blob/8900e2466e3a01c9c228ac31784aa70e8dcf3137/src/wp-includes/class-wp-customize-manager.php#L2279

### DIFF
--- a/js/customize-snapshots.js
+++ b/js/customize-snapshots.js
@@ -565,7 +565,7 @@
 			snapshot.snapshotEditContainerDisplayed.set( false );
 
 			api.state( 'snapshot-saved' ).bind( function( saved ) {
-				if ( saved ) {
+				if ( saved && ! snapshot.dirtyEditControlValues ) {
 					snapshot.updateSnapshotEditControls();
 				}
 			} );
@@ -611,6 +611,7 @@
 				}
 				snapshot.updatePending = true;
 				snapshot.editBoxAutoSaveTriggered = true;
+				snapshot.dirtyEditControlValues = false;
 				snapshot.updateSnapshot( status ).done( function() {
 					snapshot.updatePending = snapshot.dirtyEditControlValues;
 					if ( ! snapshot.updatePending ) {
@@ -621,23 +622,22 @@
 					snapshot.dirtyEditControlValues = false;
 				} ).fail( function() {
 					snapshot.updatePending = false;
+					snapshot.dirtyEditControlValues = true;
 				} );
 			}, delay );
 
 			snapshot.editControlSettings( 'title' ).bind( function() {
+				snapshot.dirtyEditControlValues = true;
 				if ( ! snapshot.updatePending ) {
 					update();
-				} else {
-					snapshot.dirtyEditControlValues = true;
 				}
 			} );
 
 			snapshot.editControlSettings( 'date' ).bind( function() {
 				if ( snapshot.isFutureDate() ) {
+					snapshot.dirtyEditControlValues = true;
 					if ( ! snapshot.updatePending ) {
 						update();
-					} else {
-						snapshot.dirtyEditControlValues = true;
 					}
 				}
 			} );
@@ -656,16 +656,13 @@
 			// @todo Show loader and disable button while auto saving.
 			api.bind( 'changeset-save', function() {
 				if ( isValidChangesetStatus() ) {
-					snapshot.updatePending = true;
 					snapshot.extendPreviewerQuery();
 				}
 			} );
 
 			api.bind( 'changeset-saved', function() {
-				api.state( 'saved' ).set( true ); // Suppress the AYS dialog.
-
-				if ( isValidChangesetStatus() ) {
-					snapshot.updatePending = false;
+				if ( 'auto-draft' === api.state( 'changesetStatus' ).get() ) {
+					api.state( 'saved' ).set( true ); // Suppress the AYS dialog.
 				}
 			});
 		},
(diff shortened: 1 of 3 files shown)


js/customize-snapshots.js
```diff
@@ -1113,6 +1113,7 @@
 				isSameStatus = api.state( 'changesetStatus' ).get() === originalOptions.data.customize_changeset_status;
 				if ( 'customize_save' === originalOptions.data.action && isSameStatus && options.data ) {
 					options.data = removeParam( options.data, 'customize_changeset_status' );
```
westonruter: Why is the `customize_changeset_status` param getting removed, again? Is this still needed?

PatelUtkarsh:
I think the code was adding status(revision) to request only when we change status from drop down.
@sayedtaqui Can you update such that it pass status when saved explicitly?

PatelUtkarsh (approved):
This looks good, Unless @sayedtaqui want to change this via using JS only.

php/class-customize-snapshot-manager.php
```diff
@@ -104,6 +104,8 @@ function hooks() {
 		add_action( 'admin_bar_menu', array( $this, 'remove_all_non_snapshot_admin_bar_links' ), 100000 );
 		add_action( 'wp_before_admin_bar_render', array( $this, 'print_admin_bar_styles' ) );
 		add_filter( 'removable_query_args', array( $this, 'filter_removable_query_args' ) );
+		add_filter( 'wp_save_post_revision_post_has_changed', array( $this, '_filter_revision_post_has_changed' ), 20, 3 );
+		add_filter( 'save_post_customize_changeset', array( $this, 'create_initial_changeset_revision' ) );
```
mohdsayed: @westonruter I think `save_post_customize_changeset ` would be an action and not filter?
https://github.com/WordPress/WordPress/blob/b0bcea2a18f095e2d5b64554782591eacf492c21/wp-includes/post.php#L3494

PatelUtkarsh:
I think this is a better approach than doing it via just JS.

js/customize-snapshots.js
mohdsayed: As we are checking `customize_snapshots_create_revision` `customize_changeset_status ` does not need to be removed now.

mohdsayed:
I think checking the param `customize_snapshots_create_revision` is smarter than removing `customize_changeset_status` plus this
```
isSameStatus = api.state( 'changesetStatus' ).get() === originalOptions.data.customize_changeset_status
```
would return true in case of existing changeset which has `draft` status so creating revision for the first time makes perfect sense 

Looks good 👍

mohdsayed (approved):

php/class-customize-snapshot-manager.php
westonruter: You're right! It should be `add_action` here.

js/customize-snapshots.js
westonruter: @sayedtaqui but core will (somewhat unintuitively) create a new revision every time the `customize_changeset_status` is supplied. Wasn't there a reason for the param to be removed? Did it have something to do with passing the date?

mohdsayed: @westonruter Yes we wanted to auto save the date and title in editbox, but we could not do `wp.customize.previewer.save()` without status because the default status is `publish`, and we did not want to create revision each time it auto saves the date/title.

westonruter: OK, so then we need to still include the logic for removing the status parameter, but then an additional change is needed right? We need to omit the request for a revision when the date and title are modified?

mohdsayed: Oh yes, thats correct.

westonruter:
@sayedtaqui I'm noticing something else here that is unexpected. If I open the customizer, change a setting, and then save the changeset as a draft; then if I click the edit icon, and then try supplying a title the changed title won't get saved until I try making another change to settings. In other words, it doesn't seem that the title changes are getting sent on `change` as expected.

Are you able to tie up the lose ends on this ticket, including sending the flag to indicate a revision should be made, as well as ensuring changes to the title cause a title update to be sent? Also, related is #112 where the date fields are being unexpectedly shown.

mohdsayed:
Sure working on them.

mohdsayed:
> try supplying a title the changed title won't get saved until I try making another change to settings. In other words, it doesn't seem that the title changes are getting sent on change as expected.

@westonruter The title would not get saved if the date is not of future, once user updates the date to be of future, it will start auto-saving, but yes it should be happening only in case of future, which I have updated. Now working on the main issue..

mohdsayed:
@westonruter  I wasn't able to finish it today, and I am going on a leave till next Tuesday, so may be @PatelUtkarsh can continue on this PR or I can do it when I come back. 👋

mohdsayed:
@westonruter Ready for review.

westonruter:
@sayedtaqui On 4.6 I'm getting an error if I Save a snapshot, then make a change, and then hit Update. A modal pops open with “The snapshot could not be saved”, with the Ajax request returning with a `bad_schedule_time` error.

PR #113, opened 2017-02-10
Add filter to button data

jonathanbardo:
While playing around the plugin I noticed it wasn't easy to change the button label or post status choices so I propose to expose a simple filter to do so.

I'm open on feedback on the filter name or if there is another way to approach this!

### DIFF
--- a/php/class-customize-snapshot-manager.php
+++ b/php/class-customize-snapshot-manager.php
@@ -666,7 +666,7 @@ class Customize_Snapshot_Manager {
 				);
 			?>
 
-			<# _.defaults( data, <?php echo wp_json_encode( $data ) ?> ); #>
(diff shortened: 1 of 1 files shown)


westonruter:
@jonathanbardo what are you intending to change the buttons to? I'm concerned about adding a filter at this point since the button UI is still in flux, per https://github.com/xwp/wp-customize-posts/issues/207

jonathanbardo:
@westonruter Sorry for the late reply, I was on vacations 😉 I think the use case would be to remove some options like "future" for example.

westonruter:
There's probably some key overlap with https://core.trac.wordpress.org/ticket/39896

There should be a filter or some core extension mechanism for the statuses that are listed.

jonathanbardo:
@westonruter I agree! We can close this for now.

Issue #101, opened 2016-11-12
Move “View on Frontend” link to customize-footer-actions; make persistent


westonruter:
In the plugin right now, a “View on Frontend” link/button appears in the `customize-header-actions` whenever the snapshot is saved. With changesets, however, the state is saved at `AUTOSAVE_INTERVAL` and it is saved whenever focus on the window is removed. In short, the changeset can be considered to be saved at any time and not require a proactive Save. Because of this, the “View on Frontend” link can be persistent and not only show up when the changeset is saved. Also, I think it makes more sense for this link to be on the bottom next to Collapse (Hide Controls) and the device preview buttons, since really this is just another way to preview the changes. Maybe it could be added along with the device preview buttons but with an [external dashicon](https://developer.wordpress.org/resource/dashicons/#external).

![image](https://cloud.githubusercontent.com/assets/134745/20235915/7cc042bc-a856-11e6-8dfd-6ffc0139cdda.png)

Here is the logic for opening the frontend URL in a new window on demand:

```js
var urlParser = document.createElement( 'a' );
urlParser.href = previewFrame.previewUrl();

params = _.extend(
	api.utils.parseQueryString( urlParser.search.substr( 1 ) ),
	{
		customize_changeset_uuid: previewFrame.query.customize_changeset_uuid,
		customize_theme: previewFrame.query.customize_theme
	}
);
urlParser.search = $.param( params );

onceProcessingComplete = function() {
	var request;
	if ( api.state( 'processing' ).get() > 0 ) {
		return;
	}

	api.state( 'processing' ).unbind( onceProcessingComplete );

	request = api.requestChangesetUpdate();
	request.done( function() {
		window.open( urlParser.href ); // <= Open the frontend URL in a new window.
	} );
};

if ( 0 === api.state( 'processing' ).get() ) {
	onceProcessingComplete();
} else {
	api.state( 'processing' ).bind( onceProcessingComplete );
}
```

westonruter:
The JS logic here should probably be encapsulated into a new core API, like `wp.customize.previewer.getFrontendPreviewUrl()`.

westonruter:
This feature has been merged into 4.9 via the Publish Settings section.

PR #115, opened 2017-02-14
Feature: Allow publishing from preview

### LINKED ISSUE #103: Consider adding Publish button to admin bar when previewing changeset
When previewing a changeset on the frontend and the Admin Bar is present, in addition to there being a Customize link to allow you to enter the customizer to make changes, there could also be a Publish button that would allow you to make the changes live right there. This would tie into a frontend editing experience for the customizer, where a user could start a frontend customizer session by clicking Customize which could then bootstrap `customize-preview.js` and start a new changeset, with inline editing of elements, floating controls, heavily relying on selective refresh (exclusively) to make changes. A user could then navigate around the site on the frontend as normal, with the changeset state persisting, and then once they are happy with the changes they can go live. Anyway, adding a Publish link/button to the admin bar is one part of this.

miina:
Resolves #103 .


js/customize-snapshots-front.js
```diff
@@ -0,0 +1,65 @@
...
+		publishBtn.click( function( e ) {
+			var request;
+			e.preventDefault();
+
+			if ( ! window.confirm( component.data.confirmationMsg ) ) { // eslint-disable-line no-alert
+				return false;
+			}
+			request = wp.ajax.post( component.data.action, {
+				nonce: component.data.snapshotsFrontendPublishNonce,
+				uuid: component.data.uuid
```
westonruter: Should this also include `theme`? Would not a theme switch also be needing to be accounted for here?

php/class-customize-snapshot-manager.php
```diff
@@ -334,6 +336,30 @@ public function enqueue_admin_scripts( $hook ) {
...
+		wp_enqueue_script( $handle );
+		$exports = array(
+			'confirmationMsg' => __( 'Are you sure that you want to publish the Snapshot?', 'customize-snapshots' ),
+			'snapshotsFrontendPublishNonce' => wp_create_nonce( 'customize-snapshots-frontend-publish' ),
+			'action' => 'customize-snapshots-frontend-publish',
+			'uuid' => $this->snapshot->uuid(),
+		);
+		wp_add_inline_script(
+			$handle,
+			sprintf( 'CustomizeSnapshotsFront.init( %s )', wp_json_encode( $exports ) ),
```
westonruter: 👍

php/class-customize-snapshot-manager.php
```diff
@@ -558,6 +589,29 @@ public function add_post_edit_screen_link( $wp_admin_bar ) {
...
+		if ( ! $this->snapshot ) {
+			return;
+		}
+		$post = $this->snapshot->post();
+		if ( ! $post ) {
+			return;
+		}
+		$wp_admin_bar->add_menu( array(
+			'id' => 'publish-customize-snapshot',
+			'title' => __( 'Publish Snapshot', 'customize-snapshots' ),
```
westonruter: Rename to “Publish Changeset”

php/class-customize-snapshot-manager.php
```diff
@@ -914,4 +973,34 @@ public function get_front_uuid_param() {
...
+		$this->ensure_customize_manager();
+		$r = $this->customize_manager->save_changeset_post( array(
+			'status' => 'publish',
+		) );
+
+		if ( is_wp_error( $r ) ) {
+			$msg = __( 'Publishing failed: ', 'customize-snapshots' );
+			foreach ( $r->errors as $name => $value ) {
+				$msg .= $name . '; ';
+			}
```
westonruter: This could be replaced with: `$msg .= join( "; ", array_keys( $r->errors ) )`. But is the it the key or the values that have the error messages to share?

php/class-customize-snapshot-manager.php
```diff
@@ -914,4 +973,34 @@ public function get_front_uuid_param() {
...
+		if ( ! check_ajax_referer( 'customize-snapshots-frontend-publish', 'nonce' ) ) {
+			status_header( 400 );
+			wp_send_json_error( 'bad_nonce' );
+		}
+
+		if ( ! isset( $_POST['uuid'] ) ) {
+			return;
+		}
+
+		$this->current_snapshot_uuid = esc_attr( $_POST['uuid'] );
```
westonruter: Replace `esc_attr( $_POST['uuid'] )` with `sanitize_key( wp_unslash( $_POST['uuid'] ) )`.

php/class-customize-snapshot-manager.php
```diff
@@ -334,6 +336,30 @@ public function enqueue_admin_scripts( $hook ) {
 	}
 
 	/**
+	 * Enqueue frontend scripts.
+	 *
+	 * These files control the behavior frontend.
+	 */
+	public function enqueue_frontend_scripts() {
+		if ( ! $this->snapshot ) {
```
westonruter: This also needs to short-circuit if the current user cannot publish changesets. In other words:

```php
if ( ! $this->snapshot || ! current_user_can( get_post_type_object( 'customize_changeset' )->cap->publish_posts ) ) {
```

php/class-customize-snapshot-manager.php
```diff
@@ -914,4 +973,34 @@ public function get_front_uuid_param() {
 	public function get_customize_uuid_param() {
 		return constant( get_class( $this->post_type ) . '::CUSTOMIZE_UUID_PARAM_NAME' );
 	}
+
+	/**
+	 * Publishes changeset from frontend.
+	 */
+	public function ajax_snapshot_frontend_publish() {
+		if ( ! check_ajax_referer( 'customize-snapshots-frontend-publish', 'nonce' ) ) {
```
westonruter: Also important to check whether `current_user_can( get_post_type_object( 'customize_changeset' )->cap->publish_posts )`

westonruter (changes requested):

php/class-customize-snapshot-manager.php
miina: Will change. Actually here we should also check for permissions I guess.

php/class-customize-snapshot-manager.php
miina: The keys seem to include the error messages.

js/customize-snapshots-front.js
miina: Added.

One thought: maybe it would be useful to include the `theme` param into the preview link in Customizer in case a non-active theme is being customized? Otherwise I change the theme, customize it, click on preview icon, and end up previewing the active theme. Thoughts?

miina: Ah, okay, just saw that this will basically be part of #101. Will pick that up.

westonruter: The only hesitation to including the `theme` in the preview link is that if a user is not authenticated they won't be able to access the preview, since previewing a theme switch requires the `switch_themes` capability.

miina:
@westonruter The changes should be done, let me know if there's anything that still needs improving.

westonruter (approved):

westonruter:
I'm waiting to do a final review and merge thus until 0.6.0 is released.

js/customize-snapshots-front.js
```diff
@@ -0,0 +1,72 @@
...
+				data = {
+					nonce: component.data.snapshotsFrontendPublishNonce,
+					uuid: component.data.uuid
+				};
+			e.preventDefault();
+
+			if ( ! window.confirm( component.data.confirmationMsg ) ) { // eslint-disable-line no-alert
+				return false;
+			}
+			if ( ! wp.customize.settings.theme.active ) {
```
miina: @westonruter After merging `develop` there is an issue with publishing changeset from preview:
screen shot 2017-07-20 at 5 53 33 pm

Looks like `wp.customize` is not defined at this moment. Do you know if anything changed meanwhile in the logic or order of enqueuing the JS scripts? Tried to make the script dependent on `customize-base` but in this case `wp.customize.settings` is undefined.

Would be great if you could point out what to investigate or what might have changed to cause this issue. Not sure what I'm missing here, maybe it's something very obvious.

westonruter: This `wp.customize.settings` value is actually set in `customize-preview` (and `customize-controls`) not `customize-base`. 

See: https://github.com/WordPress/wordpress-develop/blob/4.8.0/src/wp-includes/js/customize-preview.js#L657-L658

Note that it is set at `jQuery.ready`.

But since you'd be clicking the button after `jQuery.ready` anyway, I don't see why you it would be undefined. Is `customize-preview` itself not getting enqueued? It should be if you've loaded a changeset on the frontend.

miina: Seems like the issue was because I was using the param `changeset_uuid=...` instead of `customize_changeset_uuid` in frontend.

Note that `customize-snapshots-frontend.js` file was added to `develop` branch so I merged the frontend JS file from this branch (`customize-snapshots-front.js`) into the one in `develop` (`customize-snapshots-frontend.js`). Let me know if there are any concerns with that. Also added removing the UUID from storage after publishing.

There is another issue with publishing the changesets from frontend:
`WP_Customize_Setting::_preview_filter` is returning the dirty value as old value instead of the actual old value in DB meaning that the setting is not saved when publishing.

Although I can see that `Post_Type::hooks()` is adding this filter:
`add_action( 'transition_post_status', array( $this, 'start_pretending_customize_save_ajax_action' ), $priority - 1, 3 );`, this method is always returning true: `WP_Customize_Setting::is_current_blog_previewed()`.

Looks like `WP_Customize_Manager` is initialized in `_wp_customize_include()` before the filters have any effect. Also, not sure if https://github.com/xwp/wordpress-develop/pull/231 would fix that since the `$_REQUEST['action']` is not `customize_save` at that moment.

Am I missing something or is creating another workaround necessary, e.g. reinitializing `WP_Customize_Manager`?

westonruter:
@miina:

> Note that `customize-snapshots-frontend.js` file was added to `develop` branch so I merged the frontend JS file from this branch (`customize-snapshots-front.js`) into the one in `develop` (`customize-snapshots-frontend.js`). Let me know if there are any concerns with that. Also added removing the UUID from storage after publishing.

Yes, that's perfect.

> There is another issue with publishing the changesets from frontend:
`WP_Customize_Setting::_preview_filter` is returning the dirty value as old value instead of the actual old value in DB meaning that the setting is not saved when publishing.

If I understand right, this is the exact same problem that you've patched for core, right? In other words: https://core.trac.wordpress.org/ticket/39221

The issue is that Ajax requests on the frontend get the `customize_changeset_uuid` injected into frontend requests: https://github.com/WordPress/wordpress-develop/blob/9f3bcacd713c23ed4f71e796f46090fb6376918f/src/wp-includes/js/customize-preview.js#L373-L442

So if you can prevent this from happening via another `prefilter` handler that _removes_ the query parameter, then this would be one way to solve the problem I believe. 

Another option, perhaps a better one, would be to eliminate the Ajax request entirely in favor of just linking to an admin URL that would be responsible for publishing the changeset. In other words, it could be a `wp_nonce_url()` for which has the necessary params for publishing the post, as if the user clicked the Publish button on the edit post screen. This would be a similar nonce URL that is shown for the Trash link in the post list table. 

This would avoid having to try to override the `customize_changeset_uuid` from being added to the link, because any link in the admin bar is skipped from getting the `customize_changeset_uuid` injected into it: https://github.com/WordPress/wordpress-develop/blob/9f3bcacd713c23ed4f71e796f46090fb6376918f/src/wp-includes/js/customize-preview.js#L329-L335

miina:
@westonruter Reworked the logic to use `wp_nonce_url` instead of AJAX request, less hacky compared to pre-filtering query. Let me know if that looks something like you were thinking of.

westonruter:
@miina yes, that's it. I made a few additional changes in 4fd7832 but feel free to merge otherwise if it checks out for you.

PR #116, opened 2017-02-15
Feature: Improve frontend preview link

### LINKED ISSUE #101: Move “View on Frontend” link to customize-footer-actions; make persistent

miina:
Resolves #101


PatelUtkarsh:
@miina In a case of `auto-draft` should we add that button to preview on front-end?  Currently, It is not adding it.

I was trying to copy link but seems like you're generating it on click.

One more thing - Maybe it is not related to this PR, Currently edit post link is carrying  `customize_changeset_uuid` param. `http://vvv.dev/wp-admin/post.php?post=1946&action=edit&customize_changeset_uuid=b526cc78-d93d-431a-b027-167452915f35`

miina:
@PatelUtkarsh Thank you for testing. If you have time, could you test if you can still see these issues?

PatelUtkarsh:
@miina After publish the button stays and on click it carries `customize_changeset_uuid` however customize.php doesn't have any active/loaded changeset.

Edit: same behavior on initial load. button stays and carries `customize_changeset_uuid` param on click of frontend preview link without any change.

miina:
@PatelUtkarsh Actually it's intended to have the button static, it's like a general static preview button as the others there. The changeset uuid that it has with it is the one that's going to be added to the URL when the first change happens, basically the active but not yet visible changeset uuid. Maybe that's confusing and should be removed.

miina:
@PatelUtkarsh Should be fixed, so if you have time again ... :)

PatelUtkarsh:
@miina This looks fine except sometimes it takes time after clicking a preview button to open new tab.

miina:
It's probably taking time because it's waiting until the Customizer has finished it's changes before opening the link.

miina:
@westonruter If you have a moment could you take a quick look at this PR? To detect possible issues in code that I might have missed.

westonruter:
@miina here's a problem. I see that the preview link is getting inserted again with each time someone saves a draft and duplicates are getting inserted:

image

miina:
Thanks for bringing that out, should be fixed now.

js/customize-snapshots.js
```diff
@@ -731,16 +701,69 @@
...
+
+			snapshot.previewLink = $( $.trim( wp.template( 'snapshot-preview-link' )() ) );
+			snapshot.previewLink.attr( 'target', snapshot.data.uuid );
+			snapshot.previewLink.href = snapshot.frontendPreviewUrl.get();
+
+			snapshot.previewLink.click( function( e ) {
+				var onceProcessingComplete;
+				e.preventDefault();
+
+				snapshot.previewLink.href = snapshot.getSnapshotFrontendPreviewUrl();
```
westonruter: @miina As I was testing this, there was one use case that I found that I missed with the old version. Namely, I would often right-click on the frontend preview link in order to obtain the URL to then share with someone else or to open it in another browser. In the way that this is implemented now, this does not work. I think it should restore the previous behavior whereby there was a `setPreviewLinkHref` function that would be responsible for keeping this link up-to-date always. Then, sure, there can still be a `click` handler which is responsible for ensuring that we `requestChangesetUpdate`. But then this `requestChangesetUpdate` call could be done on `mousedown` to also capture the case where a user is copying the URL. Some combination of those events I think are needed to get it so that we can request the update both for users who try to copy the URL and for others who just want to do a simple click on the link.

westonruter (changes requested):

js/customize-snapshots.js
miina: Good finding, thanks for pointing that out, will rework the approach.

miina: @westonruter Added the same logic for updating the href as is for opening the link, do you think that's fine? (See the latest commit, f50139b).

westonruter:
There's a change of plans for where this link will appear. As can be seen in [#39896](https://core.trac.wordpress.org/ticket/39896) the link will now appear in a new panel that expands for extended publish settings:

image

So I'll be looking to incorporate your code into the core patch, specifically the JS logic. But the styling will probably not be included. And since it will be going into core directly, that will probably mean this PR will end up not getting merged.

miina:
Okay, thank you for the update.

jeffpaul:
@westonruter should we delete this PR/branch then and await the changes coming in #258 instead?

westonruter:
Yeah, we'll close the PR and will continue to refer to the code in it for the core patch in https://github.com/xwp/wordpress-develop/pull/258

js/customize-snapshots.js
```diff
@@ -731,16 +701,99 @@
...
+				params.customize_changeset_uuid = api.settings.changeset.uuid;
+			} else {
+				params.customize_changeset_uuid = snapshot.data.uuid;
+			}
+
 			a.href = snapshot.frontendPreviewUrl.get();
-			if ( a.search ) {
-				a.search += '&';
+			if ( snapshot.statusButton.disbleButton.get() ) {
+				return a.href;
```
mohdsayed, line 715: @westonruter / @miina Why are we returning just the `a.href` when the button is not enabled?

mohdsayed: Which means if the changeset is saved as draft/future/pending, we will just show the frontend without the `customize_changeset_uuid` param which will not reflect the changes on frontend.

js/customize-snapshots.js
```diff
@@ -731,16 +701,99 @@
...
+				}
+
+				api.state( 'processing' ).unbind( onceProcessingComplete );
+				request = api.requestChangesetUpdate();
+
+				request.done( function() {
+					href = snapshot.getSnapshotFrontendPreviewUrl();
+					snapshot.previewLink.attr( 'href', href );
+					if ( openLink ) {
+						window.open( href );
```
mohdsayed, line 784: And I think because of `window.open` I get popup blocked, when I click on the button. 

popup-blocked

Steps to produce: 
- Open fresh customizer without changeset uuid
- Make some changes ( I made in site title )
- Now click on the button.

mohdsayed: @westonruter

js/customize-snapshots.js
westonruter: Yes, I think that's right.

js/customize-snapshots.js
mohdsayed: This doesn't look like an ideal way to doing it. The clicking should be natural. If there are dirty changes may be let the user know through some notification like "Preparing preview.." and then let them click. Just an idea.

mohdsayed: @westonruter  Another idea would be that when you mouse enter the publish settings section, or preview link control container, we execute `api.requestChangesetUpdate()` and if there are dirty changes to be saved, make the link disabled, show a notification or loader, "Preparing preview..", and then enable the preview link when done. Thoughts?

westonruter: @sayedtaqui yes, I like that idea. If there are unsaved changes, then disable the area until you've saved them. In fact, we could actually make the appearance of the preview link conditional on you having saved changes to begin with. We wouldn't want someone try to share a link to preview an _autosaved_ state. So maybe the preview link control could appear disabled with a notification, “Save your changes in order to share the preview.” In that way you wouldn't need to have a “Preparing preview” message or any attempt to save the changeset just-in-time when clicking the link.

mohdsayed: > preview link control could appear disabled with a notification, “Save your changes in order to share the preview.”

I think this should be the best way to do it. Because if the changes are not saved as draft/future, the share link might not be useful for the person they have shared it with, because the changeset may not even exist if they have clicked on the x icon on reload or it was garbage collected ( auto-draft )

PR #117, opened 2017-02-16
Allow unauthenticated user to view a customize-draft post via REST API

### LINKED ISSUE #32: Unauthenticated user unable to view a customize-draft post via REST API
1. Create a new post via Customize Posts, giving it a `publish` status, but do not Publish
2. Click Save.
3. Make a request to via the REST API for the post, such as at http://vvv.example.com/wp-json/wp/v2/foods/159?customize_snapshot_uuid=ed6cb1bd-00c4-4927-8ae2-c98f257478f7
4. Fail: `{"code":"rest_forbidden","message":"You don't have permission to do this.","data":{"status":403}}`. It was expected that since the snapshot UUID was present, that the `publish` status in the snapshot would have allowed the post to be accessed.

The issue is in `\WP_REST_Posts_Controller::check_read_permission()`, specifically:

``` php
       // Can we read the post?
        if ( 'publish' === $post->post_status || current_user_can( $post_type->cap->read_post, $post->ID ) ) {
            return true;
        }
```

It seems the fix for this is to add `user_has_cap` filter to explicitly grant `read_post` capability for a post that is modified in the snapshot to have a status whereby the requesting user would be able to view it.

miina:
Fixes #32 

Allows unauthenticated user read a customize-draft post in case it's with status `publish` within the Snapshot/Changeset.


php/class-customize-snapshot-manager.php
```diff
@@ -914,4 +915,56 @@ public function get_front_uuid_param() {
...
+		if (
+			! $this->current_snapshot_uuid
+			|| 0 !== $current_user->ID
+			|| ! isset( $args[2] )
+			|| 'read_post' !== $args['0']
+		) {
+			return $allcaps;
+		}
+
+		$post = get_post( absint( $args[2] ) );
```
westonruter: Question: Why `absint`? Should be guaranteed to already be a post ID. If it isn't, then the previous `if` should check for it.

php/class-customize-snapshot-manager.php
```diff
@@ -914,4 +915,56 @@ public function get_front_uuid_param() {
...
+		if ( 'customize-draft' !== $post->post_status && 'auto-draft' !== $post->post_status ) {
+			return $allcaps;
+		}
+
+		// Check if the status of the post is 'published' within the changeset.
+		$chageset_id = $this->post_type->find_post( $this->current_snapshot_uuid );
+		if ( ! $chageset_id ) {
+			return $allcaps;
+		}
+		$data = $this->post_type->get_post_content( get_post( absint( $chageset_id ) ) );
```
westonruter: Mispelling in `chageset_id`.

php/class-customize-snapshot-manager.php
```diff
@@ -914,4 +915,56 @@ public function get_front_uuid_param() {
...
+
+		$is_published = false;
+		$key = 'post[' . $post->post_type . '][' . $post->ID . ']';
+
+		if ( isset( $data[ $key ] ) ) {
+			$changeset_post_values = $data[ $key ]['value'];
+			if ( isset( $changeset_post_values['post_status'] ) ) {
+				$is_published = 'publish' === $changeset_post_values['post_status'];
```
westonruter: What if the post is private and the current user can read private posts?

tests/php/test-class-customize-snapshot-manager.php
```diff
@@ -729,4 +729,45 @@ public function test_replace_customize_link() {
...
+
+		$this->manager->current_snapshot_uuid = $uuid;
+		$caps = array(
+			'0' => 'read_others_posts',
+		);
+		$args = array(
+			'0' => 'read_post',
+			'2' => $post_id,
+		);
+		$allcaps = $this->manager->filter_user_has_cap( array(), $caps, $args );
```
westonruter: Instead of using the `filter_user_has_cap` method directly, should this instead use `current_user_can` to ensure that the filter is applying?

westonruter (changes requested):

php/class-customize-snapshot-manager.php
miina: Yes, it seems to be guaranteed, good point. Added that automatically, will remove.

php/class-customize-snapshot-manager.php
miina: Unauthenticated user shouldn't have the ability to read private posts by default. Are you thinking of a case when reading private posts permission is given somewhere else to the unauthenticated user via a filter? Maybe it should be handled in the place where that's added then? Or in which cases would unauthenticated user have the permission to read private posts?

westonruter: I'm thinking of when a less-privileged authenticated user is making an API request. Consider an `editor` user making an authenticated request to the API. They should be able to see a post which is in the DB as an `draft` but which is `private` in the customized state.

php/class-customize-snapshot-manager.php
```diff
@@ -914,4 +915,56 @@ public function get_front_uuid_param() {
...
+	 * @param array $caps    Capabilities.
+	 * @param array $args    Args.
+	 * @return array All capabilities.
+	 */
+	public function filter_user_has_cap( $allcaps, $caps, $args ) {
+		global $current_user;
+
+		if (
+			! $this->current_snapshot_uuid
+			|| 0 !== $current_user->ID
```
westonruter: I think the check for current user being `0` should be removed because it should also work for under-privileged users making authenticated requests to read posts that normally they wouldn't be able to read, except that the post is made readable due to the `status` in the customized state.

Consider an author user making an authenticated API request. They should be able to access a post that is `draft` in the DB but `publish` in the customized state.

php/class-customize-snapshot-manager.php
```diff
@@ -914,4 +915,56 @@ public function get_front_uuid_param() {
...
+			|| ! isset( $args[2] )
+			|| 'read_post' !== $args['0']
```
westonruter: Why is `0` a string here?

php/class-customize-snapshot-manager.php
```diff
@@ -914,4 +915,56 @@ public function get_front_uuid_param() {
...
+		if ( isset( $data[ $key ] ) ) {
+			$changeset_post_values = $data[ $key ]['value'];
+			if ( isset( $changeset_post_values['post_status'] ) ) {
+				$is_published = 'publish' === $changeset_post_values['post_status'];
+			}
+		}
+
+		if ( true === $is_published ) {
+			$allcaps[ $caps[0] ] = true;
+		}
```
westonruter: You might want to change the condition here to be:

```php
$allcaps[ $caps[0] ] = $is_published;
```

This could potentially handle the reverse condition where access to a post could be _revoked_ if if its status is changed from `publish` to `private` and the user is not authorized to view private posts. There's probably more to it than that, but this is what came to mind.

tests/php/test-class-customize-snapshot-manager.php
```diff
@@ -729,4 +729,40 @@ public function test_replace_customize_link() {
...
+		) );
+
+		$this->manager->current_snapshot_uuid = $uuid;
+		$caps = array(
+			'0' => 'read_others_posts',
+		);
+		$args = array(
+			'0' => 'read_post',
+			'2' => $post_id,
+		);
```
westonruter: These variables, `$args` and `$caps` aren't being used?

tests/php/test-class-customize-snapshot-manager.php
```diff
@@ -729,4 +729,40 @@ public function test_replace_customize_link() {
...
+	 * Test filter_user_has_cap.
+	 *
+	 * @covers \CustomizeSnapshots\Customize_Snapshot_Manager::filter_user_has_cap()
+	 */
+	function test_filter_user_has_cap() {
+		$this->mark_incompatible();
+		$post_id = $this->factory()->post->create( array(
+			'post_type' => 'post',
+			'post_status' => 'auto-draft',
+		) );
```
westonruter, line 754: Can there be other posts here including ones that have the `draft`, `private`, and `publish` statuses? Then below in the changeset data these posts can have the `post_status` overridden to for example reverse the statuses. Then at the end it can set the current user to anonymous, to an author, and to an administrator and for each then the `read_post` cap check can be verified to return the proper result for the customized state?

miina: On it. For some reason checking for `read_post` cap causes failure (fatal error) in unit tests, haven't figured out why yet, that's why the delay in fixing.

valendesigns:
@miina Please fix merge conflicts so I can review. Thanks!

PR #118, opened 2017-02-20
Split snapshot

### LINKED ISSUE #16: Allow a snapshot to be split up into multiple snapshots
If a subset of changes from a snapshot need to be isolated into a separate snapshot, such as scheduling out a subset of settings ahead of other settings being applied, we should be able to split up a snapshot into multiple snapshots.

Given the UI for listing out the available snapshots (#12), a user could open a snapshot in an edit post screen, and from there the list of settings in the snapshot could be listed. Along with each setting there could be a checkbox to allow it to be selected, with a button then for taking the selected settings and moving them into a separate snapshot.

PatelUtkarsh:
Fixes #16


valendesigns:

PatelUtkarsh:
Todo:
- [x] test

php/class-post-type.php
```diff
@@ -1331,4 +1356,91 @@ public function handle_snapshot_fork() {
...
+		if ( ! empty( $theme ) ) {
+			$new_post_arr['meta_input']['_snapshot_theme'] = $theme;
+		}
+		$this->suspend_kses();
+		// Remove key so it doesn't save recursively.
+		unset( $_REQUEST[ $split_key ], $_POST[ $split_key ] );
+		$this->split_snapshot_id = wp_insert_post( $new_post_arr );
+		$this->split_processing_post_id = $post->ID;
+		$this->restore_kses();
+		$content = Customize_Snapshot_Manager::encode_json( $data );
```
westonruter: This needs to be wrapped in `wp_slash()`

php/class-post-type.php
```diff
@@ -424,8 +440,12 @@ public function render_data_metabox( $post ) {
 				esc_url( $frontend_view_url ),
 				esc_html__( 'Preview Changeset', 'customize-snapshots' )
 			);
-
 			echo $fork_markup; // WPCS: XSS ok.
+			if ( $should_allow_split ) {
+				$split_text = __( 'Split', 'customize-snapshots' );
+				echo sprintf( '<button id="split-activate" class="button button-secondary" data-cancel-text="%s" data-original-text="%s">%s</button>',
+				esc_html__( 'Cancel Split', 'customize-snapshots' ), esc_attr( $split_text ), esc_html( $split_text ) );
```
westonruter: Let's reformat the `sprintf()` here so that each of its args are on a separate line.

php/class-post-type.php
```diff
@@ -1331,4 +1356,91 @@ public function handle_snapshot_fork() {
...
+			$new_post_arr['meta_input']['_snapshot_theme'] = $theme;
+		}
+		$this->suspend_kses();
+		// Remove key so it doesn't save recursively.
+		unset( $_REQUEST[ $split_key ], $_POST[ $split_key ] );
+		$this->split_snapshot_id = wp_insert_post( $new_post_arr );
+		$this->split_processing_post_id = $post->ID;
+		$this->restore_kses();
+		$content = Customize_Snapshot_Manager::encode_json( $data );
+		add_action( 'post_updated', array( $this, 'redirect_split_post' ), 100 );
```
westonruter: Maybe this should happen at shutdown instead?

php/class-post-type.php
```diff
@@ -1331,4 +1356,91 @@ public function handle_snapshot_fork() {
...
+			'post_title' => $uuid,
+		);
+		$theme = get_post_meta( $post->ID, '_snapshot_theme', true );
+		if ( ! empty( $theme ) ) {
+			$new_post_arr['meta_input']['_snapshot_theme'] = $theme;
+		}
+		$this->suspend_kses();
+		// Remove key so it doesn't save recursively.
+		unset( $_REQUEST[ $split_key ], $_POST[ $split_key ] );
+		$this->split_snapshot_id = wp_insert_post( $new_post_arr );
```
westonruter: `$new_post_arr` here needs to be wrapped in `wp_slash()`

php/class-post-type.php
```diff
@@ -1331,4 +1356,91 @@ public function handle_snapshot_fork() {
...
+			'post_type' => static::SLUG,
+			'post_content' => Customize_Snapshot_Manager::encode_json( $new_split_data ),
+			'post_status' => 'draft',
+			'post_author' => get_current_user_id(),
+			'post_name' => $uuid,
+			'post_title' => $uuid,
+		);
+		$theme = get_post_meta( $post->ID, '_snapshot_theme', true );
+		if ( ! empty( $theme ) ) {
+			$new_post_arr['meta_input']['_snapshot_theme'] = $theme;
```
westonruter, line 1637: Should we include meta about the source changeset that it came from? Wouldn't this be like a fork?

PatelUtkarsh: We can add it, But where would we use it? If we are showing source from where changeset was split from that would work. Should i add that? @westonruter

PR #122, opened 2017-02-25
Fix example in doc block of wp-cli command

### LINKED ISSUE #121: Typo in dockblock
The `wp-cli` command for migration to 4.7+ in the docblock should be `snapshot` not `customize_snapshot ` here https://github.com/xwp/wp-customize-snapshots/blob/develop/php/class-customize-snapshot-command.php#L26-L27

PatelUtkarsh:
Fixes #121

### DIFF
--- a/php/class-customize-snapshot-command.php
+++ b/php/class-customize-snapshot-command.php
@@ -23,8 +23,8 @@ class Customize_Snapshot_Command {
 	 *
 	 * ## EXAMPLES
 	 *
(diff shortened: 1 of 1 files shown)


westonruter:
@PatelUtkarsh Shouldn't the WP-CLI command namespace be `snapshots` instead of `snapshot`? Or even, wouldn't `customize-snapshots` be a good choice as it aligns with the plugin slug?

PR #124, opened 2017-02-27
Fix migration notice on new WP install

### LINKED ISSUE #123: Migrate to changesets notice also shows when somebody has directly installed WordPress > 4.7
"Existing Snapshots need to be migrated to Changesets, which was added to core in WordPress 4.7." notice should not show if somebody has directly installed WordPress > 4.7, or there are no changesets to migrate.

May be we can do this here https://github.com/xwp/wp-customize-snapshots/blob/develop/php/class-migrate.php#L51-L57
```
if ( ! $this->is_migrated() ) {
    if ( empty( $this->changeset_migrate( 1, true ) ) ) {
        update_option( self::KEY, 1 );
        return;
    }
    // Add actions...
}
```

PatelUtkarsh:
Fixes #123

### DIFF
--- a/tests/php/test-class-migrate.php
+++ b/tests/php/test-class-migrate.php
@@ -108,11 +108,34 @@ class Test_Migrate extends \WP_UnitTestCase {
 	 * @see Migrate::maybe_migrate()
 	 */
 	function test_maybe_migrate() {
-		$migrate = new Migrate( $this->plugin );
+		delete_option( Migrate::KEY );
+		$migrate = $this->getMockBuilder( 'CustomizeSnapshots\Migrate' )
+		                    ->setMethods( array( 'changeset_migrate' ) )
+		                    ->setConstructorArgs( array( $this->plugin ) )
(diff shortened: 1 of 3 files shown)


tests/php/test-class-migrate.php
```diff
@@ -108,11 +108,34 @@ function test_is_migrated() {
...
+
+		// If no post to migrate fall back and add option value.
+		$migrate_obj = $this->getMockBuilder( 'CustomizeSnapshots\Migrate' )
+		                ->setMethods( array( 'changeset_migrate' ) )
+		                ->setConstructorArgs( array( $this->plugin ) )
+		                ->getMock();
+		$migrate_obj->expects( $this->once() )
+		        ->method( 'changeset_migrate' )
+		        ->with( 1, true )
+		        ->will( $this->returnValue( false ) );
```
westonruter, line 133: Fancy PHPUnit usage. 👍

PR #125, opened 2017-02-28
Fix return url in customizer.

### LINKED ISSUE #86: Published snapshot UUID can persist in the customizer exit link
Steps to reproduce:
1. Open customizer
2. Change a setting (e.g. site title)
3. Save snapshot
4. Click to view the snapshot on the frontend: 
   customize__search_results_for_ test _ _wordpress_developsss
5. On the frontend preview, click Customize in the admin bar: search_results_for_ test _ _wordpress_developsss
6. Publish changes.
7. Click ✖️ to close the customizer.

BUG: The snapshot UUID for the _published_ snapshot appears in the URL. It should be stripped when the customizer changes are published.

mohdsayed:
Fixes #86.

### DIFF
--- a/js/compat/customize-snapshots.js
+++ b/js/compat/customize-snapshots.js
@@ -48,6 +48,8 @@
 					snapshot.previewLink.attr( 'target', snapshot.data.uuid );
 				}
 
+				snapshot.removeParamFromClose( 'customize_snapshot_uuid' );
+
 				api.state( 'snapshot-exists' ).set( false );
 
 				// Replace the history state with an updated Customizer URL that does not include the Snapshot UUID.
--- a/js/customize-snapshots.js
+++ b/js/customize-snapshots.js
@@ -1123,7 +1123,7 @@
 		 * Remove a param from close button link.
 		 *
 		 * @param {string} targetParam param.
-         * @return {void}.
+		 * @return {void}.
 		 */
 		removeParamFromClose: function removeParamFromClose( targetParam ) {
 			var closeButton, queryString, params,
--- a/js/customize-snapshots.js
+++ b/js/customize-snapshots.js
@@ -203,6 +203,10 @@
 				snapshot.snapshotExpandButton.toggle( ! isPublishStatus );
 				snapshot.previewLink.toggle( ! isPublishStatus );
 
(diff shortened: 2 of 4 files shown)


westonruter (changes requested):
@sayedtaqui one more case which I noticed that needs fixing:

1. Make change in customizer.
2. Save Draft
3. Click link preview on frontend
4. Click Customize link in admin bar
5. Problem: Close button has the `customzie_changeset_uuid` query param from the frontend.

I thought that the `customize_changeset_uuid` param was being removed from the `url` param on the Customize link? Maybe it is still only removing `customize_snapshot_uuid`?

westonruter:
On a related note, if the changeset is published (or trashed) and the frontend URL is accessed with that UUID supplied, it should seem that there should also be an HTTP redirect to strip out the snapshot UUID from the URL. I suppose this would also fix the symptom for the above issue but not the underlying cause, although it would fix an issue where someone tries returning to a frontend preview URL after the changeset has been published via another means.

mohdsayed:
@westonruter You mean this https://github.com/xwp/wp-customize-snapshots/issues/86 ?

westonruter:
> You mean this #86 ?

@sayedtaqui why yes! 😄

mohdsayed:
@westonruter `removeParamFromClose()` is a little long but I thought it would be faster than using regex which I had previously written.

```js
removeParamFromClose: function removeParamFromClose( targetParam ) {
    var closeButton, returnURL, newReturnURL, UUID, initial;

    closeButton = $( '.customize-controls-close' );
    returnURL = closeButton.attr( 'href' );
    UUID = '[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}';
    initial = '\\?';

    if ( returnURL.search( new RegExp( initial + targetParam ), 'g' ) === -1 ) {
        initial = '\\&';
    }

    newReturnURL = returnURL.replace( new RegExp( initial + targetParam + '=' + UUID, 'g' ), '' );

    closeButton.attr( 'href', newReturnURL );
}
```

Ready for review.

westonruter (approved):

PR #126, opened 2017-03-02
Prevent removal of customize_changeset_status if the status are not same

mohdsayed:
In https://github.com/xwp/wp-customize-snapshots/pull/111 we had removed `isSameStatus` check before removing the `customize_changeset_status` because of which empty status was being sent when you add or change to a new status and hence throwing error for the first time in schedule. Fixed


mohdsayed:
Ready for review.

Issue #107, opened 2016-12-13
Remember the preview URL and expanded panel/section when saving a changeset


westonruter:
When a changeset is made, often the changes specifically relate to a given URL and with a specific panel, section, and control being expanded. It would be ideal if whenever a changeset is saved, that we pass along the current `previewUrl` and any expanded panel/section/control so that these params can then be used in the links generated for “Edit in Customizer” and “Preview ~~Snapshot~~Changeset” as the `url` and `autofocus` params respectively.

This closely relates to [Customizer Browser History](https://github.com/xwp/wp-customizer-browser-history), and this would greatly help with returning to work on a changeset after saving.

![image](https://cloud.githubusercontent.com/assets/134745/21152630/fec1bc0e-c11b-11e6-819a-5488ba01ea27.png)

![image](https://cloud.githubusercontent.com/assets/134745/21152730/7b2547fc-c11c-11e6-8df1-b02bdd4f3255.png)

westonruter:
Let's make this part of 0.6.0. So much for RC1 😄

mohdsayed:
Sure I am working on it. 🙂

westonruter:
Fixed by #129.

PR #129, opened 2017-03-04
Remember preview url query params

mohdsayed:
Fixes https://github.com/xwp/wp-customize-snapshots/issues/107

### DIFF
--- a/js/compat/customize-snapshots.js
+++ b/js/compat/customize-snapshots.js
@@ -1,5 +1,5 @@
 /* global jQuery, wp, _customizeSnapshotsCompatSettings, CustomizerBrowserHistory, JSON */
-/* eslint consistent-this: ["error", "snapshot"] */
+/* eslint consistent-this: ["error", "snapshot"], no-magic-numbers: [ "error", { "ignore": [0,1,2] } ] */
 
 ( function( api, $ ) {
 	'use strict';
@@ -222,13 +222,14 @@
 		 * @return {void}
 		 */
 		extendPreviewerQuery: function extendPreviewerQuery() {
-			var snapshot = this, originalQuery = api.previewer.query;
+			var snapshot = this, originalQuery = api.previewer.query, previewURLQueryParams;
 
 			api.previewer.query = function() {
 				var retval = originalQuery.apply( this, arguments );
 
-				if ( 'undefined' !== typeof CustomizerBrowserHistory ) {
-					retval.customize_preview_url_query_vars = JSON.stringify( CustomizerBrowserHistory.getQueryParams( location.href ) );
+				previewURLQueryParams = location.search.substr( 1 );
+				if ( previewURLQueryParams ) {
+					retval.customize_preview_url_query_vars = JSON.stringify( snapshot.parseQueryString( previewURLQueryParams ) );
 				}
 
 				if ( api.state( 'snapshot-exists' ).get() ) {
@@ -490,6 +491,33 @@
 
 			snapshot.updateCountdown();
 			snapshot.editContainer.find( '.reset-time' ).toggle( scheduled );
+		},
+
+		/**
+		 * Parse query string.
+		 * Taken from WP 4.7.0
+		 *
+		 * @param {string} queryString Query string.
+		 * @returns {object} Parsed query string.
+		 */
+		parseQueryString: function parseQueryString( queryString ) {
+			var queryParams = {};
+			_.each( queryString.split( '&' ), function( pair ) {
+				var parts, key, value;
+				parts = pair.split( '=', 2 );
+				if ( ! parts[0] ) {
+					return;
+				}
+				key = decodeURIComponent( parts[0].replace( /\+/g, ' ' ) );
+				key = key.replace( / /g, '_' ); // What PHP does.
+				if ( _.isUndefined( parts[1] ) ) {
+					value = null;
+				} else {
+					value = decodeURIComponent( parts[1].replace( /\+/g, ' ' ) );
+				}
+				queryParams[ key ] = value;
+			} );
+			return queryParams;
 		}
 	} );
 
--- a/js/compat/customize-snapshots.js
+++ b/js/compat/customize-snapshots.js
@@ -226,7 +226,7 @@
 
 			api.previewer.query = function() {
 				var retval = originalQuery.apply( this, arguments );
-				retval.customize_preview_url_query_vars = JSON.stringify( snapshot.getStateQueryVars() );
+				retval.customizer_state_query_vars = JSON.stringify( snapshot.getStateQueryVars() );
 
 				if ( api.state( 'snapshot-exists' ).get() ) {
 					retval.customize_snapshot_uuid = snapshot.data.uuid;
--- a/js/compat/customize-snapshots.js
+++ b/js/compat/customize-snapshots.js
@@ -222,15 +222,11 @@
 		 * @return {void}
 		 */
 		extendPreviewerQuery: function extendPreviewerQuery() {
-			var snapshot = this, originalQuery = api.previewer.query, previewURLQueryParams;
+			var snapshot = this, originalQuery = api.previewer.query;
 
 			api.previewer.query = function() {
 				var retval = originalQuery.apply( this, arguments );
-
-				previewURLQueryParams = location.search.substr( 1 );
-				if ( previewURLQueryParams ) {
-					retval.customize_preview_url_query_vars = JSON.stringify( snapshot.parseQueryString( previewURLQueryParams ) );
-				}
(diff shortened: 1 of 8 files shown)


mohdsayed:
Ready for review.

 I am not sure if `preg_match( '/[a-z|\[|\]|_|-|0-9]+/', $value )` would be the correct choice for validating the panel/section/control id , please suggest

js/compat/customize-snapshots.js
```diff
@@ -226,6 +226,11 @@
 
 			api.previewer.query = function() {
 				var retval = originalQuery.apply( this, arguments );
+
+				if ( 'undefined' !== typeof CustomizerBrowserHistory ) {
+					retval.customize_preview_url_query_vars = JSON.stringify( CustomizerBrowserHistory.getQueryParams( location.href ) );
```
westonruter: I don't think that Customizer Browser History should be made a dependency for this feature.

This can be replaced with:

```js
retval.customize_preview_url_query_vars = wp.customize.utils.parseQueryString( location.search.substr( 1 ) );
```

though on 4.6 a polyfill for `wp.customize.utils.parseQueryString` would be needed.

js/customize-snapshots.js
```diff
@@ -966,6 +966,11 @@
 
 			api.previewer.query = function() {
 				var retval = originalQuery.apply( this, arguments );
+
+				if ( 'undefined' !== typeof CustomizerBrowserHistory ) {
+					retval.customize_preview_url_query_vars = JSON.stringify( CustomizerBrowserHistory.getQueryParams( location.href ) );
```
westonruter: See above.

php/class-customize-snapshot-manager.php
```diff
@@ -934,4 +943,73 @@ public function get_front_uuid_param() {
...
+
+		$allowed_query_params = array_merge( $allowed_panel_section_control_params, array(
+			'device',
+			'scroll',
+		) );
+
+		$allowed_devices = array(
+			'mobile',
+			'desktop',
+			'tablet',
```
westonruter: This list needn't be hard-coded. You can use `\WP_Customize_Manager::get_previewable_devices()`.

php/class-customize-snapshot-manager.php
```diff
@@ -934,4 +943,73 @@ public function get_front_uuid_param() {
...
+		$preview_url_query_vars = array();
+		$params = wp_array_slice_assoc( $original_query_vars, $allowed_query_params );
+
+		if ( ! empty( $params ) ) {
+			foreach ( $params as $param => $value ) {
+				$is_valid_var = (
+					(
+						in_array( $param, $allowed_panel_section_control_params )
+						&&
+						preg_match( '/[a-z|\[|\]|_|-|0-9]+/', $value )
```
westonruter: This regex seems to be missing the initial `^(` and ending `)$` if it is doing what I think.

php/class-customize-snapshot-manager.php
```diff
@@ -104,7 +104,8 @@ function hooks() {
 		add_action( 'admin_bar_menu', array( $this, 'remove_all_non_snapshot_admin_bar_links' ), 100000 );
 		add_action( 'wp_before_admin_bar_render', array( $this, 'print_admin_bar_styles' ) );
 		add_filter( 'removable_query_args', array( $this, 'filter_removable_query_args' ) );
-		add_action( 'save_post_customize_changeset', array( $this, 'create_initial_changeset_revision' ) );
+		add_action( 'save_post_' . $this->get_post_type(), array( $this, 'create_initial_changeset_revision' ) );
+		add_action( 'save_post_' . $this->get_post_type() , array( $this, 'save_customize_preview_url_query_vars' ) );
```
westonruter: There's an extra space before the comment here.

php/class-post-type.php
```diff
@@ -849,4 +862,21 @@ public function filter_out_settings_if_removed_in_metabox( $content ) {
...
+	 * Get preview url query vars.
+	 *
+	 * @param int $post_id Post id.
+	 * @return array $preview_url_query_vars Preview url query vars.
+	 */
+	public function get_preview_url_query_vars( $post_id ) {
+		$preview_url_query_vars = array();
+
+		// If customizer browser history plugin is active.
+		if ( function_exists( 'customizer_browser_history_enqueue_scripts' ) && $post_id ) {
```
westonruter: Per above, I think Customizer Browser History shouldn't be made a dependency.

westonruter (changes requested):

php/class-customize-snapshot-manager.php
mohdsayed: The panel/sections/control ids can be like
`post[post][1437]`
`title_tagline`
`blogname`
`nav_menu-61`
`postmeta[posttype][27450][keywords]`

It doesn't follow any initial or ending pattern?

westonruter: Isn't the purpose to validate that the `$param` is a valid ID string? If you want to ensure that _only_ the valid characters are included then you should do:

```php
preg_match( '/^[a-z|\[|\]|_|-|0-9]+$/', $value )
```

Adding the `^` and `$` anchors, as otherwise a user could supply an ID like: `!@#$%^&*()title_tagline⁄€‹›ﬁﬂ‡°·‚`

php/class-post-type.php
```diff
@@ -849,4 +862,20 @@ public function filter_out_settings_if_removed_in_metabox( $content ) {
...
+	 * Get preview url query vars.
+	 *
+	 * @param int $post_id Post id.
+	 * @return array $preview_url_query_vars Preview url query vars.
+	 */
+	public function get_preview_url_query_vars( $post_id ) {
+		$preview_url_query_vars = array();
+
+		if ( $post_id ) {
+			$preview_url_query_vars = (array) get_post_meta( $post_id, '_preview_url_query_vars', true );
```
westonruter: If there is no postmeta on the changeset post, where `get_post_meta()` returns `false`, this will result in `$preview_url_query_vars === array( false )`. 

So this should probably be:

```php
$preview_url_query_vars = get_post_meta( $post_id, '_preview_url_query_vars', true );
if ( ! is_array( $preview_url_query_vars ) ) {
    $preview_url_query_vars = array();
}
```

westonruter (changes requested):

php/class-customize-snapshot-manager.php
mohdsayed: @westonruter I think I am not able to get it, Can you see this http://regexr.com/3ffv4 ?

mohdsayed: I should probably be `\[a-z|\[|\]|_|\-|0-9]+\` ( escaped - sign ) but `/^[a-z|\[|\]|_|-|0-9]+$/` doesn't match anything.

westonruter: @sayedtaqui from that example, you can see that it is matching all of those strings, the last one incorrect.

Compare:

image

and 

image

mohdsayed: got it.

mohdsayed:
`Class 'PHPUnit_Framework_TestCase' not found in /tmp/wordpress/tests/phpunit/includes/testcase. `

Not sure why the build is failing.

westonruter:
> Not sure why the build is failing.

@sayedtaqui it looks like an issue with PHPUnit 6.0. I'm noticing the same thing locally after upgrading from 5.x.

See also https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/870#issuecomment-285260433

See also:
> Looks like travis-ci just updated its PHP 7 images to use PHPUnit 6+, so our PHP 7 builds are failing. We should either fast-track #39822 or install + use and older version of PHPUnit in our travis script
https://wordpress.slack.com/archives/core/p1489073667330796

php/class-customize-snapshot-manager.php
```diff
@@ -104,7 +104,8 @@ function hooks() {
 		add_action( 'admin_bar_menu', array( $this, 'remove_all_non_snapshot_admin_bar_links' ), 100000 );
 		add_action( 'wp_before_admin_bar_render', array( $this, 'print_admin_bar_styles' ) );
 		add_filter( 'removable_query_args', array( $this, 'filter_removable_query_args' ) );
-		add_action( 'save_post_customize_changeset', array( $this, 'create_initial_changeset_revision' ) );
+		add_action( 'save_post_' . $this->get_post_type(), array( $this, 'create_initial_changeset_revision' ) );
```
westonruter, line 107: Good catch!

westonruter (approved):

westonruter:
@sayedtaqui Ready for merge after you give a final review of my tweaks.

mohdsayed:
👍  Lets merge.

Issue #131, opened 2017-03-11
Publish button not enabled on theme switch


westonruter:
I tried previewing a different theme and the Publish button was not enabled so I couldn't activate the new theme unless I first dirtied a setting:

![image](https://cloud.githubusercontent.com/assets/134745/23820887/d5010090-05d7-11e7-8c58-307f556b5e20.png)

PR #132, opened 2017-03-11
Enable status button if current theme is not active.

mohdsayed:
Fixes https://github.com/xwp/wp-customize-snapshots/issues/131

### DIFF
--- a/js/customize-snapshots.js
+++ b/js/customize-snapshots.js
@@ -120,7 +120,7 @@
 			queryVars.url = api.previewer.previewUrl.get();
 
 			if ( ! api.state( 'activated' ).get() || snapshot.data.theme !== currentTheme ) {
-				queryVars.previewingTheme = true;
+				queryVars.previewing_theme = true;
 			}
 
 			_.find( [ 'control', 'section', 'panel' ], function( constructType ) {
@@ -314,7 +314,7 @@
 		 * @return {void}
 		 */
 		addButtons: function addButtons() {
-			var snapshot = this, disableButton = true,
+			var snapshot = this, disableButton = true, disableSelectButton = true,
 				setPreviewLinkHref, currentTheme;
 
 			currentTheme = api.settings.theme.stylesheet;
@@ -325,6 +325,7 @@
 			snapshot.statusButton = snapshot.addStatusButton();
 
 			if ( api.state( 'changesetStatus' ).get() ) {
+				disableSelectButton = false;
 				if ( 'auto-draft' === api.state( 'changesetStatus' ).get() ) {
 					disableButton = false;
 				} else {
@@ -334,9 +335,11 @@
 
 			if ( ! api.state( 'activated' ).get() || snapshot.data.theme !== currentTheme ) {
 				disableButton = false;
+				disableSelectButton = false;
 			}
 
-			snapshot.statusButton.disable( disableButton );
+			snapshot.statusButton.disbleButton.set( disableButton );
+			snapshot.statusButton.disableSelect.set( disableSelectButton );
 
 			// Preview link.
 			snapshot.previewLink = $( $.trim( wp.template( 'snapshot-preview-link' )() ) );
--- a/js/customize-snapshots.js
+++ b/js/customize-snapshots.js
@@ -19,6 +19,7 @@
 			initialServerDate: '',
 			initialServerTimestamp: 0,
 			initialClientTimestamp: 0,
+			theme: '',
 			i18n: {},
 			dirty: false
 		},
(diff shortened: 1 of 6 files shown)


mohdsayed:
@westonruter We have a param `theme=twentyfifteen` in url when we switch theme, I think that should also be saved in changeset like we saved the history query params in the other PR?

westonruter:
@sayedtaqui oh yes, absolutely! That param can be included in the customizer state query params only when saving a changeset for a theme that is not active. If the theme is active, then the param should be omitted from being saved.

westonruter:
Is this out of WIP and ready for review?

js/customize-snapshots.js
```diff
@@ -104,15 +107,22 @@
...
 				'autofocus[section]': null,
 				'autofocus[panel]': null
 			};
+			currentTheme = api.settings.theme.stylesheet;
 			queryVars.scroll = parseInt( api.previewer.scroll, 10 ) || 0;
 			queryVars.device = api.previewedDevice.get();
 			queryVars.url = api.previewer.previewUrl.get();
 
+			if ( ! api.state( 'activated' ).get() || snapshot.data.theme !== currentTheme ) {
+				queryVars.previewingTheme = true;
```
westonruter: Since this is going to be sent with the POST request, I think it should be `previewing_theme` instead. Then below in PHP, you'd read ` $query_vars['previewing_theme']`.

php/class-post-type.php
```diff
@@ -917,11 +920,17 @@ public function get_frontend_view_link( $post ) {
...
-			),
-			$base_url
+		$current_theme = get_stylesheet();
+		$args = array(
+			static::FRONT_UUID_PARAM_NAME => $post->post_name,
 		);
+
+		if ( isset( $preview_url_query_vars['theme'] ) && $current_theme !== $preview_url_query_vars['theme'] ) {
+			$args = array_merge( $args, array(
+					'theme' => $preview_url_query_vars['theme'],
```
westonruter: Extra tab here?

westonruter (changes requested):

mohdsayed:
Not ready yet, need to test and may be few changes. Trying to finish today.

mohdsayed:
Ready for review.

js/customize-snapshots.js
```diff
@@ -255,6 +266,7 @@
 					dialogElement;
 
 				snapshot.statusButton.disableSelect.set( false );
+				snapshot.statusButton.disbleButton.set( false );
```
westonruter, line 269: I'm glad you caught this. It's something I was noticing in the case of setting validation errors, where I'd get a validation error message with a control but the Publish button would remain disabled.

westonruter:
@sayedtaqui the one remaining issue I see, and this isn't strictly related to the changes here, but when I publish the changeset the frontend preview link continues to have the `customize_changeset_uuid` even though it is published:

image

php/class-customize-snapshot-manager-back-compat.php
```diff
@@ -106,7 +101,7 @@ public function enqueue_controls_scripts() {
 			'initialServerDate' => current_time( 'mysql', false ),
 			'initialServerTimestamp' => floor( microtime( true ) * 1000 ),
 			'theme' => $this->original_stylesheet,
-			'previewingTheme' => isset( $preview_url_query_vars['theme'] ) ? $preview_url_query_vars['theme']: '',
+			'previewingTheme' => ! $this->customize_manager->is_theme_active(),
```
mohdsayed: @westonruter This is breaking my condition here https://github.com/xwp/wp-customize-snapshots/blob/9ca3ff2f90df27da5fc62946b7857b61dbcc3184/js/compat/customize-snapshots.js#L278
There are several use cases where the button has to be enabled or disabled.

1.  When they switch theme ( Has to enable ) ( Working )
2. After saving the previewing theme, reload the page, ( Has to disable ) ( Not working now )
3. After saving the previewing theme they switch to a third theme ( Has to enable ) ( Working )
4. After travelling through the themes and saving the last one, they come back to the original theme ( Should be enable ) ( Not working now ).

I was not able to satisfy these conditions with only boolean value of `previewingTheme `. This is as good as `api.state( 'activated' ).get()`

westonruter: Thanks for pointing this out. So this can be fixed by reverting back to:

```php
'previewingTheme' => isset( $preview_url_query_vars['theme'] ) ? $preview_url_query_vars['theme'] : '',
```
?

mohdsayed: Yes.

php/class-customize-snapshot-manager.php
```diff
@@ -225,7 +226,7 @@ public function add_snapshot_uuid_to_return_url() {
 			&&
 			$this->current_snapshot_uuid
 			&&
-			$this->is_theme_active()
+			$this->customize_manager->is_theme_active()
 			&&
```
mohdsayed, line 230: @westonruter Not sure why `$this->customize_manager->is_theme_active()` is returning `false` in 4.6.1 which is failing phpunit test.  I mean inside phpunit test.

westonruter: It's because the `setup_theme` action wasn't triggered in the test case, and so `\WP_Customize_Manager::$original_stylesheet` didn't get set. In 4.7 this variable gets set in `\WP_Customize_Manager::__construct()` whereas in 4.6 it got set in `\WP_Customize_Manager::setup_theme()`.

westonruter: Should be fixed in 52694b693.

westonruter (approved):

PR #134, opened 2017-04-17
Fix date handling compatibility with Safari

westonruter:
I opened the customizer in Safari and I found:

> [Error] TypeError: null is not an object (evaluating 'parsed[fieldName]') at (anonymous function) (customize-snapshots.js:774)

It looks like some logic from Customize Posts was adopted before fixes for its Safari issues were implemented. See:
https://github.com/xwp/wp-customize-posts/pull/293
https://github.com/xwp/wp-customize-posts/pull/304

### DIFF
--- a/js/customize-snapshots.js
+++ b/js/customize-snapshots.js
@@ -771,7 +771,7 @@
 				snapshot.schedule.inputs.each( function() {
 					var input = $( this ),
 						fieldName = input.data( 'date-input' );
-					$( this ).val( parsed[fieldName] );
+					$( this ).val( parsed[ fieldName ] );
 				} );
 			}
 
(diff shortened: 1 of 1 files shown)


js/customize-snapshots.js
```diff
@@ -886,12 +886,12 @@
...
-			formattedDate += '-' + ( '00' + date.getDate() ).substr( -nonYearLength, nonYearLength );
-			formattedDate += ' ' + ( '00' + date.getHours() ).substr( -nonYearLength, nonYearLength );
-			formattedDate += ':' + ( '00' + date.getMinutes() ).substr( -nonYearLength, nonYearLength );
-			formattedDate += ':' + ( '00' + date.getSeconds() ).substr( -nonYearLength, nonYearLength );
+			formattedDate = ( '0000' + String( date.getFullYear() ) ).substr( -yearLength, yearLength );
+			formattedDate += '-' + ( '00' + String( date.getMonth() + monthOffset ) ).substr( -nonYearLength, nonYearLength );
+			formattedDate += '-' + ( '00' + String( date.getDate() ) ).substr( -nonYearLength, nonYearLength );
+			formattedDate += ' ' + ( '00' + String( date.getHours() ) ).substr( -nonYearLength, nonYearLength );
+			formattedDate += ':' + ( '00' + String( date.getMinutes() ) ).substr( -nonYearLength, nonYearLength );
+			formattedDate += ':' + ( '00' + String( date.getSeconds() ) ).substr( -nonYearLength, nonYearLength );
```
westonruter, line 894: Being explicit about types just in case.

js/customize-snapshots.js
```diff
@@ -980,22 +980,32 @@
...
+		 *
+		 * @param {string} postDate Post date string.
+		 * @returns {Date} Parsed date.
+		 */
+		parsePostDate: function parsePostDate( postDate ) {
+			var dateParts = _.map( postDate.split( /\D/ ), function( datePart ) {
+				return parseInt( datePart, 10 );
+			} );
+			return new Date( dateParts[0], dateParts[1] - 1, dateParts[2], dateParts[3], dateParts[4], dateParts[5] ); // eslint-disable-line no-magic-numbers
+		},
```
mohdsayed, line 1003: I wish we could pass an array in the `Date` object, but one day we will be able to do it using spread syntax, like this, ( though it works in chrome ) 😀
```
return new Date(...dateParts);
```

js/customize-snapshots.js
```diff
@@ -980,22 +980,32 @@
 		 * @returns {string} Current datetime string.
 		 */
 		getCurrentTime: function getCurrentTime() {
-			var snapshot = this,
-				currentDate = new Date( snapshot.data.initialServerDate ),
-				currentTimestamp = snapshot.dateValueOf(),
-				timestampDifferential;
-
+			var snapshot = this, currentDate, currentTimestamp, timestampDifferential;
+			currentTimestamp = ( new Date() ).valueOf();
```
mohdsayed, line 984: Just for my knowledge, thinking why you chose `( new Date() ).valueOf()`  over `( new Date() ).getTime();`

mohdsayed (approved):

js/customize-snapshots.js
westonruter: Good question. I don't have a particular reason. I just usually use it, as has the same output: 

> You can use this method to help assign a date and time to another Date object. This method is functionally equivalent to the valueOf() method.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTime

I agree that `getTime()` would perhaps look more semantic.

PR #160, opened 2017-08-29
Add required PHP version go readme

### LINKED ISSUE #159: Add minimum PHP version to readme
See https://twitter.com/TacoVerdo/status/901000997383876609

westonruter:
Fixes #159.

See https://make.wordpress.org/plugins/2017/08/29/minimum-php-version-requirement/

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -10,6 +10,7 @@ Provide a UI for managing Customizer changesets; save changesets as named drafts
 **Tested up to:** 4.8.1  
 **Stable tag:** 0.6.2  
 **License:** [GPLv2 or later](http://www.gnu.org/licenses/gpl-2.0.html)  
+**Requires PHP:** 5.3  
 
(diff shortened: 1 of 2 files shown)


valendesigns:
@westonruter Shouldn't this be updated in the readme generator in dev-lib?

westonruter:
@valendesigns updated in what way?

westonruter:
The readme generator is copying “Requires PHP” into the markdown version.

valendesigns:
Add `Requires PHP` to this array https://github.com/xwp/wp-dev-lib/blob/master/class-wordpress-readme-parser.php#L41 and inject it into the generated readme.md

westonruter:
It doesn't seem to be required because it works without it.

valendesigns:
I see, so the readme generator is automatically adding `Requires PHP` even though it's not mentioned in the file anywhere. Awesome!

westonruter:
I think the predefined list in the script just ensures that the same order is used when generated.

PR #161, opened 2017-09-24
[WIP] 4.9 Refactoring Test

mohdsayed:
Test branch just to check compatibility with the new changes in 4.9.

PR #162, opened 2017-10-21
Upgrade for WP 4.9 and remove back-compat for <4.7

westonruter:
- [x] Conditionally incorporate https://github.com/xwp/wp-customize-snapshots/pull/161 when on 4.9+.
- [x] Make sure features not merged into core are ported into `customize-snapshots.js`, including changeset title and saving the current previewed URL and expanded panels/sections as meta when saving a changeset, etc.
- [x] Fix autosave requests. They currently fail with a `changeset_post_save_failure` for `edit_others_posts`. There is some conflict between setting the caps in `\CustomizeSnapshots\Post_Type::extend_changeset_post_type_object()` and the logic to in `\WP_Customize_Manager::grant_edit_post_capability_for_changeset()` when performing an autosave request.
- [x] Update changeset title when user sitting on customizer and schedule time arrived.
- [ ] Test and fix how pending state is saved when user does not have publish capability.

**Testing:**
- [x] Test if able to publish changeset from front end in compat and 4.9
- [x] Test if able to merge changesets in compat and 4.9
- [x] Test if snapshot admin post edit screen is working as expected in compat and 4.9
- [x] Test state query vars are saved properly in compat and 4.9
- [x] Test if able to create multiple changesets in compat and 4.9
- [x] Overall testing


westonruter:
@sayedtaqui the autosave problem actually seems to be super easy to fix. See 71e4bc1.

js/customize-snapshots.js
```diff
@@ -1,104 +1,121 @@
...
+				api.state.create( 'changesetInspectLink', snapshot.data.inspectLink );
 
-				if ( api.state.has( 'changesetStatus' ) && api.state( 'changesetStatus' ).get() ) {
-					api.state( 'snapshot-exists' ).set( true );
-				}
+				api.control( 'changeset_scheduled_date', snapshot.setupScheduledChangesetCountdown );
 
-				snapshot.extendPreviewerQuery();
+				api.bind( 'save-request-params', function( data ) {
+					if ( api.state( 'selectedChangesetStatus' ) && 'publish' !== api.state( 'selectedChangesetStatus' ) ) {
```
westonruter: The `publish` check doesn't seem right here. The title should be sent regardless of the status.

js/customize-snapshots.js
```diff
@@ -1,104 +1,121 @@
...
-						snapshot.spinner.removeClass( 'is-active' );
+			api.control.add( titleControl );
 
-						$( id ).dialog( {
-							autoOpen: true,
-							modal: true
-						} );
-					}
-				} );
+			toggleControl = function( status ) {
```
westonruter: This can be removed. The title field should always be shown. I should be able to open the Customizer, make a change, provide a title (commit message), and then publish.

js/customize-snapshots.js
```diff
@@ -1,104 +1,121 @@
...
-				api.state.create( 'snapshot-submitted', true );
-
-				snapshot.data.uuid = snapshot.data.uuid || api.settings.changeset.uuid;
-				snapshot.data.title = snapshot.data.title || snapshot.data.uuid;
+				var saveBtn = $( '#save' );
 
-				snapshot.editBoxAutoSaveTriggered = false;
+				snapshot.data.title = snapshot.data.title || api.settings.changeset.uuid;
+				api.state.create( 'changesetTitle', snapshot.data.title );
+				api.state.create( 'changesetInspectLink', snapshot.data.inspectLink );
```
westonruter: Maybe `changesetInspectUrl` would be better instead of link?

js/customize-snapshots.js
```diff
@@ -139,1090 +156,104 @@
...
-					if ( snapshot.isFutureDate() ) {
-						snapshot.countdown.show();
-						snapshot.updateSnapshot( status );
-					}
-				} else {
-					snapshot.updateSnapshot( status );
-					snapshot.snapshotEditContainerDisplayed.set( false );
-					snapshot.countdown.hide();
+				toggleEditLinkControl: function() {
+					this.active.set( _.contains( [ 'draft', 'future', 'pending' ], api.state( 'changesetStatus' ).get() ) );
```
westonruter: I think the condition here should be opposite. Namely, it should be deactivated if the current status is `''` or `'auto-draft'`. This would be more future-proof if other plugins add their own statuses.

js/customize-snapshots.js
```diff
@@ -139,1090 +156,104 @@
...
-
-			removeParam = function( queryString, parameter ) {
-				var pars = queryString.split( /[&;]/g );
-
-				_.each( pars, function( string, index ) {
-					if ( string && string.lastIndexOf( parameter, 0 ) !== -1 ) {
-						pars.splice( index, 1 );
-					}
+		addPendingToStatusControl: function() {
+			var snapshot = this, params, coreStatusControl, index = 2;
```
westonruter: Instead of hard-coding the index of `2`, why not find the index dynamically? If it is supposed to be after `draft`, then find the draft and add one?

php/class-post-type-back-compat.php
```diff
@@ -1,241 +0,0 @@
...
-		if ( $maybe_error ) {
-			$states['snapshot_error'] = __( 'Error on publish', 'customize-snapshots' );
-		}
-		return $states;
-	}
-
-	/**
-	 * Show an admin notice when publishing fails and the post gets kicked back to pending.
-	 */
-	public function show_publish_error_admin_notice() {
```
westonruter, line 224: I think this method needs to be restored.

php/class-post-type-back-compat.php
```diff
@@ -1,241 +0,0 @@
...
-
-	/**
-	 * Display snapshot save error on post list table.
-	 *
-	 * @param array    $states Display states.
-	 * @param \WP_Post $post   Post object.
-	 *
-	 * @return mixed
-	 */
-	public function display_post_states( $states, $post ) {
```
westonruter, line 210: Restore?

php/class-customize-snapshot-manager.php
```diff
@@ -410,6 +286,8 @@ public function create_initial_changeset_revision( $post_id ) {
 	/**
 	 * Prepare snapshot post content for publishing.
 	 *
+	 * @todo Move to compat?
```
westonruter: I don't think so. The idea for the `publish_error` is that if you have a changeset saved as a draft, but then you add a new validation constraint or remove a setting from being registered, that if you then try to publish the changeset from the admin, that the changeset will get blocked from publishing and an error message will then be displayed when editing the post.

westonruter: Humm. I can see that it will be somewhat challenging to restore this `publish_error` functionality in 4.9 because `save_settings_with_publish_snapshot` in the plugin is now `\WP_Customize_Manager::_publish_changeset_values` in core. We'd have to somehow suspend the `_wp_customize_publish_changeset` function from running during the `transition_post_status` action when it detects that there is an invalid setting.

For example:

```php
add_action( 'transition_post_status', function( $new_status, $old_status, $changeset_post ) {
   // @todo Check if there are any invalid settings in $changeset_post...
   $has_invalid_settings = true; // ...

   // Prevent publishing of settings.
   if ( $has_invalid_settings ) {
        remove_action( 'transition_post_status', '_wp_customize_publish_changeset' );
        add_action( 'transition_post_status', function( $n, $o, $post ) {
            if ( $post->ID === $changeset_post->ID ) {
                add_action( 'transition_post_status', '_wp_customize_publish_changeset' );
            }
        }, 100, 3 ); // later
    }
}, 9, 3 ); // Before _wp_customize_publish_changeset() at priority 10.

````

mohdsayed: If its not very important right now, can I do this in a separate PR, I spent some time but could't get my head around this very well, I need to look back and forth to better understand, and since `show_publish_error_admin_notice`  and `display_post_states` were kept compat `php/class-post-type-back-compat.php`, so I assume its not being used for 4.7 anyways, though I have restored them.

westonruter: Yes, we can restore that functionality later. Normally it should not be used. It's more of an edge case.

westonruter:
@sayedtaqui anything left here as far as you know?

mohdsayed:
I think this is it. I will do more testing before you wake up.

PR #171, opened 2017-11-24
Add workaround to fix pending status causing infinite loop when user does not have publish capability

### LINKED ISSUE #167: Clicking pending status freezes the browser when user does not have publish capability.
In the new 0.7.0 version, when user does not have publish changeset capability, and only draft and pending appears in status control in publish settings section, clicking the pending option freezes the browser.

mohdsayed:
Workaround for https://github.com/xwp/wp-customize-snapshots/issues/167

Fixes #167.

### DIFF
--- a/js/customize-snapshots.js
+++ b/js/customize-snapshots.js
@@ -1,6 +1,7 @@
 /* global wp, jQuery */
 /* eslint consistent-this: [ "error", "snapshot", "control" ] */
 /* eslint no-magic-numbers: ["error", { "ignore": [0, 1] }] */
+/* eslint max-nested-callbacks: [ "error", 4 ] */
 
 (function( api, $ ) {
 	'use strict';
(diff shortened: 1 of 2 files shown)


php/class-customize-snapshot-manager.php
```diff
@@ -202,6 +202,7 @@ public function enqueue_controls_scripts() {
 		// Script data array.
 		$exports = apply_filters( 'customize_snapshots_export_data', array(
 			'inspectLink' => isset( $edit_link ) ? $edit_link : '',
+			'addWorkaroundFor42686' => version_compare( strtok( get_bloginfo( 'version' ), '-' ), '4.9', '==' ),
```
westonruter: We should also make sure the version is under `4.9.2` as soon as we get the fix committed to core.

js/customize-snapshots.js
```diff
@@ -53,6 +54,15 @@
 					snapshot.addPendingToStatusControl();
 					snapshot.addTitleControl( section );
 					snapshot.addInspectChangesetControl( section );
+
+					/**
+					 * This is not ideal but is just a workaround to fix https://core.trac.wordpress.org/ticket/42686 for 4.9.0.
+					 */
+					section.expanded.bind( function( expanded ) {
```
westonruter: Why only upon `expanded`?

mohdsayed: I wanted to change it as late as possible so that any other thing is not affected because of this change. Like the button text should not be publish
https://github.com/xwp/wordpress-develop/blob/master/src/wp-admin/js/customize-controls.js#L7944-L7948

Also upon `expanded`,  because the user will change the status only when he expands the publish section , and that's the time when this `api.settings.changeset.currentUserCanPublish` needs to be `true`

